Issue #30, opened 2013-08-21
no Keypress detection during split or raw modes


ghost:
Hello,
reading the source code (select.c display.c and split.c) 
I can see some keypress functions but they doesn't work,
inside select.c at line 183 the switch block is never tested,
the surrounding if is always FALSE...
I tried it on FreeBSD and Linux putting some printf.
any ideas to further investigate the key press events?
thank you

PR #51, opened 2014-07-12
Disable keyboard input in "split" mode

zorun:
Keyboard input does not seem useful in split mode, as the output is
probably meant to be parsed by another program.  Users can press ^C anyway
if they want to interrupt mtr.

It was helplessly broken anyway: no key press was actually taken into
account (see bug #30 on Github), _except_ the "Enter" key, which would
cause an infinite loop by calling select() repeatedly.

### DIFF
--- a/display.c
+++ b/display.c
@@ -58,7 +58,6 @@ extern int DisplayMode;
 #define split_open()
 #define split_close()
 #define split_redraw()
-#define split_keyaction() 0
 #else
 #include "split.h"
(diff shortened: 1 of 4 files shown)


cvidler:
zorus, thank you this was driving me crazy.  bash script I was writing would work in any other mode than --split, but I wanted the --split output format.

I wasn't getting a random infinite loop, it'd do it every execution.

PR #65, opened 2014-12-12
Ignore the space field in XML report.

dmbaturin:
Space field doesn't make much sense in XML report anyway, and would make it ill-formed if used. Since it is used in the default fields spec ("LS NABWV"), it makes the default report ill-formed.

Also, for some reason it produces something I wouldn't really expect:

```
    <HUB COUNT="7" HOST="google-public-dns-a.google.com">
        <Loss>  0.0%</Loss>
        <Snt>    10</Snt>
        < > </(null)>
        <Last>  80.2</Last>
        <Avg>  80.3</Avg>
        <Best>  80.1</Best>
        <Wrst>  80.4</Wrst>
        <StDev>   0.0</StDev>
    </HUB>
```
I'm not sure why the closing tag contains NULL. Maybe my dirty hack is for the wrong problem, or there is another problem that I don't see yet.
This behaviour existed before my previous patch about the Loss% field.

### DIFF
--- a/report.c
+++ b/report.c
@@ -294,7 +294,7 @@ void xml_close(void)
     printf("    <HUB COUNT=\"%d\" HOST=\"%s\">\n", at+1, name);
     for( i=0; i<MAXFLD; i++ ) {
       j = fld_index[fld_active[i]];
-      if (j < 0) continue;
+      if (j <= 0) continue;
 
       strcpy(name, "        <%s>");
       strcat(name, data_fields[j].format);


dmbaturin:
It's been a while. Any comments about it?

rewolff:
I would say a simple patch would be to not issue the "space" field if encountered. 

I don't like "special cases" but I'm afraid we'll have to add one here.

dmbaturin:
That is, expressly ignore the space field when doing XML output? I'll look into it, closing this one.

rewolff:
Hey... The "closed pull request" alerted me that a patch was included. 
The <  becomes <= is elegant, but requires a comment. Patch applied.

dmbaturin:
Ah, yeah, it was a pull request, not an issue. :)
Thanks, I think I can drop my forked repo now, until I want to add/fix something else.
Hope distro maintainers will pick it up soon.

The reason I bother is that I want to write a kind of network/system unit testing tool that will consists of a test interpreter and scripts that pass an environment in JSON or XML to it. It would be nice to have a script for traceroute so tests can check path length or that traffic goes through specific nexthop, but standard traceroute output is not very machine friendly, while MTR XML output, when it's well-formed, is just what I need.

Issue #42, opened 2014-02-28
Feature request: Set display mode from command line


eik3:
That would be great!

WoLpH:
+1 on this, as a temporary workaround I currently use this script to start mtr in a tmux session:

```
#!/bin/sh

test -z "$base_session" && exit 1

tmux -2 new-session -d -s $base_session -n mtr

tmux send-keys -t $base_session:0 'mtr --timeout 2 8.8.8.8' C-m
sleep 0.5
tmux send-keys -t $base_session:0 'dd'
tmux select-window -t $base_session:0
tmux -2 attach-session -t $base_session
```

PR #70, opened 2015-01-05
Added --displaymode option

kiwano:
It's been bugging me for a while that mtr always starts in stats mode, and lacks a command line switch to start in one of the stripchart modes, so I added the switch --displaymode, and included it in the man page and usage output. (This resolves issue #42 btw)


joshuadfranklin:
This is awesome! Any chance you could do a release so this starts showing up downstream?

PR #81, opened 2015-06-05
Add SCTP support

swordfeng:
Simply in a TCP-like way.


swordfeng:
There may be some bugs, and sorry that I have missed some other files that need to be modified...

PR #82, opened 2015-06-05
SCTP support

swordfeng:
I added codes about SCTP in *curses.c* and *select.c* this time, and got SIGPIPE when MTRing to a non-existent address (which causes a loop route between two routers).
I just modified *net.c*, to avoid the SIGPIPE, but I don't know if I'm doing the right thing, as I'm not familiar with the whole program.

### DIFF
--- a/curses.c
+++ b/curses.c
@@ -264,6 +264,7 @@ int mtr_curses_keyaction(void)
     switch ( mtrtype ) {
     case IPPROTO_ICMP:
     case IPPROTO_TCP:
+    case IPPROTO_SCTP:
       mtrtype = IPPROTO_UDP;
       break;
     case IPPROTO_UDP:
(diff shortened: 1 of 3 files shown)


LionNatsu:
How about the bug now?

LionNatsu:
ref.:  #81

swordfeng:
@LionNatsu 
I did not dig deeply into the code... It seems that the previous PR just works well.
I'll be happy if you decide to do some further work.

curses.c
```diff
@@ -276,12 +277,25 @@ int mtr_curses_keyaction(void)
...
+    case IPPROTO_ICMP:
+    case IPPROTO_UDP:
+    case IPPROTO_TCP:
+      mtrtype = IPPROTO_SCTP;
+      break;
+    case IPPROTO_SCTP:
+      mtrtype = IPPROTO_ICMP;
+      break;
+    }
     return ActionNone;
```
LionNatsu: new feature: here's a new switch key `s` for SCTP

select.c
```diff
@@ -107,7 +107,7 @@ void select_loop(void) {
     FD_SET(netfd, &readfd);
     if(netfd >= maxfd) maxfd = netfd + 1;
 
-    if (mtrtype == IPPROTO_TCP)
+    if (mtrtype == IPPROTO_TCP || mtrtype == IPPROTO_SCTP)
```
LionNatsu: add the event handler(s) for socket? (I am not good at operating file descriptor)

curses.c
LionNatsu: Oh, it will surely not work. `s` conflicted with line 135, the key to `Change Packet Size`.

LionNatsu:
BTW, why does TCP not emit SIGPIPE?

PR #91, opened 2015-10-01
Some improvements to the documentation.

garethrandall:


rewolff:
Thanks for the reminder. I think the preferred way to report bugs is to use this github bug tracking system. Can you make an additional patch suggesting that? (I'll merge this because it's already better than referring to the no longer existing mailing list.)

PR #93, opened 2015-10-12
curses: Fix background transparency in terminal

### LINKED ISSUE #72: After adding colors in ncurses, transparency is gone
Sorry for submitting the issue too early, I hit enter unintentionally. =)

So. After, I guess, f2bc4aac0d135a99f3bcf4bed65f62676459d87a, and package update in ArchLinux (0.85→0.86, December 25th), `mtr` window is no longer transparent.

I guess, the background color is explicitly set to black.

Screenshot:

screenshot mtr no transparency

Above, my `TERM=screen`. When run with `TERM=linux`, e.g., it looks like this:

screenshot mtr no transparency

If there’s some env variable I could set to have the colors off, I’d rather have transparent and monochrome output. Of course, colored one *with* transparency would be the cutest. ♥

Narthorn:
Properly use "-1" as terminal background color, when use_default_colors() succeeds.

Patch comes from, and closes #72.

### DIFF
--- a/curses.c
+++ b/curses.c
@@ -700,13 +700,15 @@ void mtr_curses_open(void)
   initscr();
   raw();
   noecho(); 
+  int bg_col = 0;
   start_color();
 #ifdef HAVE_USE_DEFAULT_COLORS
-  use_default_colors();
+  if (use_default_colors() == OK)
+    bg_col = -1;
 #endif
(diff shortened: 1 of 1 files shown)


michalrus:
Thank you! :heart:

PR #98, opened 2015-11-12
State that Github is the preferred way to report bugs.

garethrandall:
As you requested, I've updated the README to state that bugs should be submitted to the Github bug tracking system.


rewolff:
Thanks!

PR #103, opened 2016-01-04
Fix typos

jwilk:


rewolff:
Thanks. This sort of stuff needs doing every now and then!

PR #104, opened 2016-02-23
updated net.c to 12 to fix cloud network timeouts

wadem:
Updated MAX_UNKNOWN_HOSTS to 12, from 5. Have been hearing of issues where -f is becoming default/required in some larger cloud networks. This value allows mtr to work in all tested cases again without -f and gives head room for additional network growth.

### DIFF
--- a/net.c
+++ b/net.c
@@ -151,8 +151,9 @@ struct sequence {
 
 /* Configuration parameter: How many queries to unknown hosts do we
    send? (This limits the amount of traffic generated if a host is not
-   reachable) */
-#define MAX_UNKNOWN_HOSTS 5
+   reachable)  Adjusted from 5 to 12. Some cloud networks appear to have 
+   greater than 5 hosts before resuming. */
+#define MAX_UNKNOWN_HOSTS 12
 
 
 /* BSD-derived kernels use host byte order for the IP length and 


LionNatsu:
This may close because the feature has been implemented in a later commit https://github.com/traviscross/mtr/commit/a8532c78b68e69b98a9c22d714eae0f381dade5d by @yupeng820921.

wadem:
I disagree. --max-unknown option in commit a8532c7 is still set to 5, which is the problem. This default value needs to be higher to support cloud provider networks and their growth. Else when clouds are adding additional network devices MTR is starting to break in same cases due to their internal hops. Through testing with various cloud providers, 12 is a safe number to ensure we don't hit this issue for many more years to come. It adds minimal overhead on the existing infrastructure.

LionNatsu:
You're right. I agree with you.
But this pull-request is also broken (cannot merge).

LionNatsu:
I wrote a new PR for you just now, is it okay?

rewolff:
As commented in the new PR, I don't like to increase the network load too much for the standard case where A) hops respond normally, and B) the routes are not that long.

wadem:
@LionNatsu thanks for updating the PR - appreciated.

wadem:
For tracking, this PR was merged in PR#130 (https://github.com/traviscross/mtr/pull/130)

PR #120, opened 2016-07-03
add max-unknown option

yupeng820921:
remove the MAX_UNKNOWN_HOSTS macro, use maxUnknown variable
instead of it, and add --max-unknown option, let it can be
configured at run time.

I hit the limit of MAX_UNKNOWN_HOSTS, my test path has 6 unknown host, So I should recompiled mtr and in crease the value of MAX_UNKNOWN_HOSTS. I hope mtr can support to adjust this value at run time.

### DIFF
--- a/mtr.8
+++ b/mtr.8
@@ -80,6 +80,9 @@ mtr \- a network diagnostic tool
 .BI \-m \ MAX\-TTL\c
 ]
 [\c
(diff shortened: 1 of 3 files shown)


rewolff:
You have one tabs/spaces mixup. (all others were correct, I think).

PR #124, opened 2016-07-21
Fix GTK display order.  Update an email address in the GTK credits display.

aquerubin:
Avg and Best were swapped in the GTK display.

Update my email address in the GTK credits.

### DIFF
--- a/gtk.c
+++ b/gtk.c
@@ -409,7 +409,7 @@ void TreeViewCreate(void)
   g_object_set (G_OBJECT(renderer), "xalign", 1.0, NULL);
   column = gtk_tree_view_column_new_with_attributes ("Snt",
     renderer,
-    "text", 3,
+    "text", COL_SNT,
     "foreground", COL_COLOR,
     NULL);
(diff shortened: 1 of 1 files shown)


rewolff:
Tony... Second opinion: How about: I tag a version number on the current master  (i.e. release it). 
Then. we merge newdns into master? Then because nobody runs bleeding edge, tag a version number on that version as well. That should get newdns some testing, as well as allow us to say: well revert to version XX if you're having trouble.

aquerubin:
On Thu, 21 Jul 2016, rewolff wrote:

> Tony... Second opinion: How about: I tag a version number on the current 
> master (i.e. release it). Then. we merge newdns into master? Then 
> because nobody runs bleeding edge, tag a version number on that version 
> as well. That should get newdns some testing, as well as allow us to 
> say: well revert to version XX if you're having trouble.

Yep.  Works for me.

Antonio Querubin
e-mail:  tony@lavanauts.org
xmpp:  antonioquerubin@gmail.com

PR #126, opened 2016-07-29
add option to set graceperiod

russor:
useful when running a single probe on a network where many hops never respond


rewolff:
Looks clean. :-)

russor:
Thanks, had this lying around and saw you were preparing to do a release, so I figured I should get it in :)

PR #127, opened 2016-07-29
Add support for JSON as report output format

nikolaik:
This is based on the equivalent code for XML. It should produce valid JSON (with proper commas and everything). The new command line option is --json.

### DIFF
--- a/display.c
+++ b/display.c
@@ -88,6 +88,9 @@ void display_open(void)
   case DisplayTXT:
     txt_open();
     break;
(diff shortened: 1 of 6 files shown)


rewolff:
hihi. Fun! I'll  Look into it later.

leleobhz:
-z and -e options does not reflect in generated JSON. Its normal?

LionNatsu:
@leleobhz Ummm... It should acts like an XML output...

PR #130, opened 2016-08-03
Change default --max-unknown from 5 to 12 for large cloud networks

LionNatsu:
Fix the PR #104 to fit the new implementation. 

> updated net.c to 12 to fix cloud network timeouts
>
> Updated MAX_UNKNOWN_HOSTS to 12, from 5. Have been hearing of issues where -f is becoming default/required in some larger cloud networks. This value allows mtr to work in all tested cases again without -f and gives head room for additional network growth.
> 
> ...This default value needs to be higher to support cloud provider networks and their growth. Else when clouds are adding additional network devices MTR is starting to break in same cases due to their internal hops. Through testing with various cloud providers, 12 is a safe number to ensure we don't hit this issue for many more years to come. It adds minimal overhead on the existing infrastructure.

### DIFF
--- a/mtr.c
+++ b/mtr.c
@@ -89,8 +89,8 @@ int  fstTTL = 1;                /* default start at first hop */
 /*int maxTTL = MaxHost-1;  */     /* max you can go is 255 hops */
 int   maxTTL = 30;              /* inline with traceroute */
                                 /* end ttl window stuff. */
-int maxUnknown = 5;				/* stop send package */
-                                /*when larger than this count */
+int maxUnknown = 12;            /* stop send package */
+                                /* when larger than this count */
 int remoteport = 0;            /* for TCP tracing */
 int localport = 0;             /* for UDP tracing */
 int tcp_timeout = 10 * 1000000;     /* for TCP tracing */


LionNatsu:
CC: @wadem

rewolff:
I'm not sure I agree with this. You are seeing more than five consecutive hosts that ignore TCPIP RFCs and refuse to send any error messages when things go awry. You can add -f 12 to your personal MTR_OPTIONS if you want to.

LionNatsu:
Hum....okay so, @wadem, should I leave this open for any further discussion? or close it?

wadem:
I speak from experience that cloud providers like GCP and AWS are moving to networks where this is becoming more common @rewolff , yes. 5 is too small for the growing cloud infrastructure. It's hard to say how much time is being wasted by their support companies having to teach end users to use the -f flag. This is a real problem, 5 is too small. I came across this issue (and fix) as a result of hitting it. I was told to work around the problem, and be one of the many with a cloud provider hitting this. I would rather this is fixed at the source (here) instead of working around it. This is why I created the original PR. @LionNatsu thanks for re-merging the PR, I had not been keeping track.

wadem:
Thank you

LionNatsu:
Fixed.

PR #133, opened 2016-08-13
In UDP mode multiple concurrent mtr's can get mixed up

jrouvier-neumob:
UDP mode uses a source port of the current PID (unless a specific source port has been specified).  This, however, is not checked when receiving packets.  As a result two or more concurrent mtr's can get mixed up.  This change aborts processing if the returned packet doesn't have the expected source port.

**Test case:**
Run two mtr's, the first to a very close host, such as your default gateway, the other mtr to something far away.  Start the 2nd mtr quicky after the first mtr.  For example, if 192.168.1.1 was your default gateway and you are 100ms+ away from health.gov.za:

`mtr -c10 --report -u 192.168.1.1 > mtr_near & sleep 0.1 ; mtr -c10 --report -u health.gov.za > mtr_far`

before this patch, both `mtr_near` and `mtr_near` will contain the route to `health.gov.za`

### DIFF
--- a/net.c
+++ b/net.c
@@ -1001,6 +1001,15 @@ void net_process_return(void)
         break;
 #endif
       }
+
+      if (!localport) {
+        if (ntohs(udpheader->srcport) != (uint16)getpid())
+          return;
+      } else {
+        if (ntohs(udpheader->srcport) != (uint16)localport)
+          return;
+      }
+
       if (remoteport && remoteport == ntohs(udpheader->dstport)) {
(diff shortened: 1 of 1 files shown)


rewolff:
I've fixed it in a different way. My patch leaves the codebase smaller.... 
Can you check my fix?  I tried your "demo case" (with the old binary), but do not see your problem.

aquerubin:
I'm able to duplicate the problem here.  Your patch works.

rewolff:
OK.

PR #134, opened 2016-08-14
new usage() and few cleanups

kerolasa:
This pull request is mostly about making usage output useful. Earlier printout of short options without explanation did not give enough context to understand what the options might do without looking to manual page, and that renders the help output useless. The new output is hopefully much more helpful, altough it requires more maintenance care.

Rest of the changes are clean ups of various sorts.

### DIFF
--- a/asn.c
+++ b/asn.c
@@ -157,23 +157,23 @@ char* trimsep(char *s) {
 }
 
 // originX.asn.cymru.com txtrec:    ASN | Route | Country | Registry | Allocated
-char* split_txtrec(char *txtrec) {
-    if (!txtrec)
+char* split_txtrec(char *txt_rec) {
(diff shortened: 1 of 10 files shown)


rewolff:
Although "struct names" and "variable names" live in a separate namespace, I don't like having overlaps. 
This goes for "struct hostent" and the variable named "hostent" after your patch. 

I don't feel strong enough about this to reject the patch for that nitpick.... if you were to fix it with another patch/pull, it will be accepted. If it doesn't happen it probably won't get "fixed". Oh well.... 

e.g. the index -> strchr change was slightly on my todo list. (I spotted an "index" wondered what that did, read the documentation (deprecated) and ... didn't change the others. ) Thanks for putting in the legwork! :-)

PR #135, opened 2016-08-15
Warnings branch

kerolasa:
The second pull request is mostly about fixing all the remaining warnings I could find with clang, gcc, and cppcheck. Now when the build is clean I can start focusing to bulid-sys change set. For example 'make distcheck' does not work, but that shall be addressed in the next pull.

p.s. I normally have worked with project using maillist for change reviews, so please be patient while I try to get up to speed with github interface.

### DIFF
--- a/curses.c
+++ b/curses.c
@@ -518,9 +518,6 @@ void mtr_curses_init() {
 
 	/* Initialize block_map. */
 	block_split = (NUM_FACTORS - 2) / 2;
-	if (block_split > 9) {
-		block_split = 9;
-	}
 	for (i = 1; i <= block_split; i++) {
 		block_map[i] = '0' + i;
 	}
--- a/net.c
+++ b/net.c
@@ -865,7 +865,7 @@ void net_process_return(void)
   struct sockaddr * fromsockaddr = (struct sockaddr *) &fromsockaddr_struct;
   struct sockaddr_in * fsa4 = (struct sockaddr_in *) &fromsockaddr_struct;
   socklen_t fromsockaddrsize;
-  int num;
+  ssize_t num;
   struct ICMPHeader *header = NULL;
   struct UDPHeader *udpheader = NULL;
   struct TCPHeader *tcpheader = NULL;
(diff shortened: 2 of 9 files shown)


report.c
```diff
@@ -472,7 +472,7 @@ void csv_close(time_t now)
 
       /* 1000.0 is a temporay hack for stats usec to ms, impacted net_loss. */
       if( strchr( data_fields[j].format, 'f' ) ) {
-	printf( ",%.2f", data_fields[j].net_xxx(at) / 1000.0);
```
rewolff: This one doesn't make sense.  in the original the division upcasts the "data_fields" to double and is then divided. The double is passed to printf. 

In your new version, the division still has the implicit upcast, the result is downgraded to float and then again upgraded to double to pass to printf. 
(remember floating point arguments are always passed as doubles.  in C). 

Moving the closing parenthesis (you added) forward a bit would be better....

net.c
```diff
@@ -865,7 +865,7 @@ void net_process_return(void)
   struct sockaddr * fromsockaddr = (struct sockaddr *) &fromsockaddr_struct;
   struct sockaddr_in * fsa4 = (struct sockaddr_in *) &fromsockaddr_struct;
   socklen_t fromsockaddrsize;
-  int num;
```
rewolff: Would it clean up further, do you think, if num would be "size_t", and that you cast just when doing the recvfrom?

curses.c
```diff
@@ -518,9 +518,6 @@ void mtr_curses_init() {
 
 	/* Initialize block_map. */
 	block_split = (NUM_FACTORS - 2) / 2;
-	if (block_split > 9) {
```
rewolff: The original (with the constant probably unnecessary) meant to convey that > 9 is not allowed for the following code. Maybe add a comment to this effect?

rewolff:
@kerolasa, I've placed a few commens in the "patch". 

I like what you're doing, and to prevent slowing you down, I'm going to accept this pull request, but I'd appreciate it if you could go back and look at those again. Possibly change those with my suggestion in a later patch......

PR #136, opened 2016-08-16
2016 08 16 make build system better

kerolasa:
This pull is mostly about making the autotools stuff nicer.  One of the commits, e2d898cc16398609821f7a786ae86dcbac4884af, is a bit of a monster to be reviewed. Perhaps it is easiest not to read ever single line but to observe that ./bootstrap && ./configure && make still works. 

Rest of the changes are straightforward, and at the beginning of the pull there are two commits e2d898cc16398609821f7a786ae86dcbac4884af and 78f019aa9d49be5ada89296a462939738c20c181 that should address comments from previous pull.

Now I am at cross roads, and I think I should open a bug item whether this project wants to be maxium portable or use linux facilities to maxium. Let me do that next.


report.c
```diff
@@ -472,7 +472,7 @@ void csv_close(time_t now)
 
       /* 1000.0 is a temporay hack for stats usec to ms, impacted net_loss. */
       if( strchr( data_fields[j].format, 'f' ) ) {
-	printf( ",%.2f", (float) (data_fields[j].net_xxx(at) / 1000.0));
```
rewolff: Apparently I wasnt clear. 
`someint/1000.0
`promotes someint to a double before doing the calculation. The result is a double. Why would you cast that to float or double before passing it as an argument to printf?
you could make the "upgrade" of the integer explicit by saying:
`((double)someint) / 1000.0
`and in the case at hand because the someint is complex, you don't want the first part of that expression be casted but the whole thing. So it would become
`((double) (data_fi.....(at))) / 1000.0`

report.c
```diff
@@ -49,19 +48,15 @@ extern int af;
...
-  char *t;
-  now = time (NULL);
-  t = ctime (&now);
-  t [ strlen (t) -1] = 0; // remove the trailing newline
-  return t;
-}
-
 void report_open(void)
 {
-  printf ("Start: %s\n", get_time_string ());
```
rewolff: I like the old version better here. 
The reason is that you can read "report_open" and understand what it is trying to do just by reading the words in the function. 
Similarly, the function "get_time_string" does just what its name implies. 

So... in two places the "get_time_string" DOCUMENTS what the code is doing, even though it is not efficient and if you're looking at the details, you get an extra indirection.

rewolff:
Again two nitpicks, but the majority is good.... You know the drill. :-)

rewolff:
You broke the build on my system. :-(
bootstrap, configure --without-gtk 
results in 

```
(CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /home/wolff/mtr/build-aux/missing autoheader)
autoheader: warning: missing template: NO_GTK
autoheader: Use AC_DEFINE([NO_GTK], [], [Description])
make: *** [config.h.in] Error 1
```
(I HAVE built this on this system the last few weeks a few times. It is an ubuntu 14.04 system. it doesn't have gtk installed. )

rewolff:
I tried git bisect for the first time. It points the finger at: 
e2d898c... build-sys: improve configure.am

rewolff:
Can you check that 467ca18a5 is a reasonable fix? I am a complete newbie at "autoconf" and friends.... After over 20 years of maintaining mtr....

PR #138, opened 2016-08-17
2016 08 17 regression fixes and warnings removal

kerolasa:
Sorry about bit of a mess with autotools revamp. Here are fixes to mine and others regressions. In same go some warnings got to be removed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -102,7 +102,7 @@ AC_ARG_WITH([ipinfo],
   [AS_HELP_STRING([--without-ipinfo], [Do not try to use ipinfo lookup at all])],
   [], [with_ipinfo=yes])
 AM_CONDITIONAL([IPINFO], [test "x$with_ipinfo" = "xyes"])
-AS_IF([test "x$ipinfo" = "xno"], [
+AS_IF([test "x$with_ipinfo" = "xno"], [
   AC_DEFINE([NO_IPINFO], [1], [Define to disable ipinfo lookup])
 ])
 
@@ -113,7 +113,7 @@ AC_ARG_ENABLE([ipv6],
 USES_IPV6=
 AC_CHECK_FUNC([getaddrinfo], [
   AS_IF([test "x$WANTS_IPV6" = "xyes"], [
-    AC_DEFINE([ENABLE_IPV6], [], [Define to enable IPv6])
+    AC_DEFINE([ENABLE_IPV6], [1], [Define to enable IPv6])
     USES_IPV6=yes
   ])
 ])
--- a/curses.c
+++ b/curses.c
@@ -505,7 +505,7 @@ void mtr_gen_scale(void)
 
 static char block_map[NUM_FACTORS];
 
-void mtr_curses_init() {
+void mtr_curses_init(void) {
 	int i;
 	int block_split;
 
--- a/dns.c
+++ b/dns.c
@@ -49,7 +49,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 //#include <ctype.h>
-//#include <string.h>
+#include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <signal.h>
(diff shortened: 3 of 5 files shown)


aquerubin:
On Wed, 17 Aug 2016, Sami Kerola wrote:

> Sorry about bit of a mess with autotools revamp. Here are fixes to mine 
> and others regressions. In same go some warnings got to be removed. You 
> can view, comment on, or merge this pull request online at:
>
>  https://github.com/traviscross/mtr/pull/138
>
> -- Commit Summary --
>
>  * warnigns: add void to functions that do not take any arguments
>  * build-sys: fix --without-ipinfo regressions
>  * build-sys: fix ./configure --disable-ipv6
>  * warnings: fix --disable-ipv6 --without-ipinfo compilation warnings
>
> -- File Changes --
>
>    M configure.ac (4)
>    M curses.c (2)
>    M dns.c (10)
>    M net.c (28)
>    M report.c (2)
>
> -- Patch Links --
>
> https://github.com/traviscross/mtr/pull/138.patch
> https://github.com/traviscross/mtr/pull/138.diff~.

In the interest of simplifying the codebase, I think all this effort to 
disable IPv6 at build-time should be redirected to detecting whether IPv6 
is enabled/useable at run-time and then doing the right thing.  I suspect 
there are few if any current POSIX-compliant environments that don't have 
support for the API functions that specifically provide IPv6 capability.

Many of the problems we've had in the past with IPv6 had more to do with 
the custom resolver code failing when IPv6 is either not configured or 
explicitly disabled in the OS.  But that part of the code is now legacy.

Antonio Querubin
e-mail:  tony@lavanauts.org
xmpp:  antonioquerubin@gmail.com

rewolff:
Antonio, I agree that "detecting not-working IPV6 at turntime" should work. On the other hand, I don't want to lose compile-ability on systems that don't have the interface.

rewolff:
kerolasa, I'm going to do the UNUSED things a bit different. I'm going to pull and then fix a few of your UNUSED things another way... I'm not sure if my compiler supports the unused keyword, so I'm not sure if I will see the difference.

rewolff:
kerolasa, it seems that you put the "UNUSED" after the declaration of the argument: 
`void somefunc (int arg UNUSED)
`
while other places have the UNUSED before the argument: 
`void somefunc (UNUSED int arg)
`
I take it both are allowed, it seems I'm not getting any errors/warnings.

aquerubin:
On Wed, 17 Aug 2016, rewolff wrote:

> Antonio, I agree that "detecting not-working IPV6 at turntime" should 
> work. On the other hand, I don't want to lose compile-ability on systems 
> that don't have the interface.

I fully support maintaining compile-ability but my question though is what 
reasonably current systems do NOT support the API?  Ie. it doesn't make 
sense to sprinkle the code with a bunch of ifdefs for code that would not 
cause a build error on a current system or even one that's 10+ years old. 
If you do find such an old system, I would then claim there are probably 
other far more important issues with the continued use of that OS than the 
ability to rebuild the latest and greatest version of mtr for it.  Sure 
you may need ifdefs for some of the actual code as a substitute for lack 
of runtime logic but for things like variable declarations do you really 
think it's necessarry?


Antonio Querubin
e-mail:  tony@lavanauts.org
xmpp:  antonioquerubin@gmail.com

rewolff:
I dislike the ifdefs... I would like them outside the main code path. 
for example: 
```
#ifdef ENABLE_IPV6
  struct sockaddr_storage name_struct;
#else
  struct sockaddr_in name_struct;
#endif
```
This should have been fixed with a
```
#ifdef ENABLE_IPV6
#define mtr_sockaddr_storage sockaddr_storage
#else
#define mtr_sockaddr_storage sockaddr_in
#endif
```
but in net.c there is lots of code "extra" for ipv6. If we can make that compile into "do nothing" code at compiletime under an `#ifndef ENABLE_IPV6` then that would clean things up a lot.

Issue #139, opened 2016-08-18
standard deviation


aquerubin:
There seems to be a significant change in the displayed standard deviation values in curses mode.  This may have begun with commit bc3972.  In many cases the RTT for a particular hop will change value a number of times as mtr progresses while the StDev value remains unchanged at 0.  Running with the previous commit generates values that appear to make sense.  

Running in gtk mode shows yet another range of standard deviation values altogether.

LionNatsu:
A suggestion: at least 7 numbers for a commit hash

rewolff:
I personally don't think std deviation numbers are a good measure for most MTR numbers. If you have a long route, and all packets take say 130ms give or take a little, then the standard deviation is a valid measure. But in most cases, your packets will take 130ms in 95% of the cases, and MUCH longer the remaining cases. This results in a big number for stddev. 

IMHO a more interesting parameter would be "slow packets percentage". Minimum +20% would be a naive threshold. but it depends on the link. I now have an example trace: 163.2 161.8 172.6 (avg, best, worst over 55 pings)  where a value above 175 would need to be called "slow". Same trace, a hop closer to home: 11.7   1.4 118.2 where average and stdev are "unfairly" dominated  by a few slow packets.

aquerubin:
On Thu, 18 Aug 2016, rewolff wrote:

> I personally don't think std deviation numbers are a good measure for 
> most MTR numbers. If you have a long route, and all packets take say 
> 130ms give or take a little, then the standard deviation is a valid 
> measure. But in most cases, your packets will take 130ms in 95% of the 
> cases, and MUCH longer the remaining cases. This results in a big number 
> for stddev.

Variation in the RTT is an important measure of jitter.  If the large 
majority of the packets are all 1000ms with a few outliers +/- 10ms the 
std dev would be expected to be low particularly for long-running mtr. 
But where you've got say just the first 5 cycles and the RTTs are spread 
evenly between say 10 ms and 15 ms, the StDev for that should NOT still be 
0 after the first 5 cycles.  And then compare that curses output with the 
gtk output.  Something is broken in the StDev calculation.

Antonio Querubin
e-mail:  tony@lavanauts.org
xmpp:  antonioquerubin@gmail.com

rewolff:
Agreed. Patch welcome. (As mentioned, I don't care much about stddev. And I don't know the proper formulas by heart.).

aquerubin:
On Thu, 18 Aug 2016, rewolff wrote:

[...]

PR #140, opened 2016-08-18
Fix standard deviation calculation.

aquerubin:
Fixes issue #139


rewolff:
Thanks! Even tough I don't use it much, it looks more plausible now.

PR #142, opened 2016-08-21
Pkg config

kerolasa:
This pull request relates almost entirely to build system.  Change 262291a makes some of the oldest and most broken operating environments not to be able to compile this projec, but I doubt such systems would have worked earlier.

Secondly the pkg-config is made explicitly required. Again some users on some ancient systems may be upset about this, but IMHO the pkg-config, portable and future proved way to find dependencies in uniformed expected manner. And it is not even that controversial anymore - see https://en.wikipedia.org/wiki/GNU_Build_System#Components that has pkg-config listed along side with make, gettext, and gcc.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,7 +19,7 @@ mtr_SOURCES = mtr.c mtr.h \
               img/mtr_icon.xpm \
               mtr-gtk.h
 
-if IPINFO
+if WITH_IPINFO
 mtr_SOURCES += asn.c asn.h
 endif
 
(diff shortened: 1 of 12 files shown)


rewolff:
Thank you for this patch! 
Some autoconf / pkg-config magic seems to work! 
I ran bootstrap on three different systems and was able to configure and make on those three systems wihtout any problems!

yvs2014:
Also it can affect on mtr building in some systems without pkg-config, gtk, and so on..
% uname -srm; which pkg-config
FreeBSD 10.2-RELEASE-p14 i386
pkg-config not found

rewolff:
@kerolasa  can you look for a solution for this?

PR #143, opened 2016-08-21
2016 08 21

kerolasa:
First one of the changes makes user  input validation more strict. Second is about improving error messaging. Rest are for people who look the code. One magic constant removal, the getopt go-around compilation is made conditional to avoid local code when libc can be used, and finally I decided ot help compiler to generate better optimizations by informing what functions are not needed in global name space.


rewolff:
Looks good. 
1)  typo "defalt" instead of "default"  (I will fix this). 
2) You need to add a autoconf check for the "error" function. It is marked: 
> These functions and variables are GNU extensions, and should not be
>        used in programs intended to be portable.
So to keep with the portability goals: We write code that uses the modern interfaces (i.e. call error), but then provide a compatibility function.

PR #144, opened 2016-08-23
Mac fixes

kerolasa:
I got my hands on mac, and found couple portability issues that did not exist before I started changing stuff. This pull request addresses these issues. That said it would be nice to get all relevant systems checked before next release if they still work. And if not there is always the option of Big Hammer and integrate with gnulib.


rewolff:
I would prefer to have an 
`#include "portability/mtr_error.h"`
where that one does
```
#if HAVE_ERROR
#include <error.h>
#else 
... rest of file
#endif 
```

PR #146, opened 2016-08-29
Datatypes

kerolasa:
Primary objective in this pull request is to clean up data types from global scope to function scope. This is needed to achieve cleaner api. Some of the changes are a bit too large, but I did not see how to split them any further, appologies about that. And as usual there are few changes that are not strictly speaking under subject of data type clean up, but they are probably fine.

Please review crash fix commit carefully. To be honest I don't understand what was tried with %5N notation, so my change might be totatlly wrong.


kerolasa:
These changes will be part of pull 147.

PR #147, opened 2016-09-04
data types, various cleanups, clarifications, corrections, and few fixes

kerolasa:
This is likely a pull request in both number of commits and amount of lines changed this project has never seen, and hopefully will never have to see in future. I don't know how review could happen effectively, but perhaps looking each change once and trying that all works at after checking out commit 7a8648607dbaa1dd678cfba4d19017263c0261c3 is reasonable way to go.

Assuming these changes are nice, good, and there is no need for immediate other changes I reckon making a release v0.88 would be most appropriate sometime soon. Notice that I am saying that from only my behalf, perhaps there are other changes that will need work before tagging a release is  reasonable -  I am simply too new to this project to be aware what else might be going.


LionNatsu:
(Umm...Split it into several minor PRs may be easier to review...)

rewolff:
I reviewed the top 5% and the bottom 10% of the "files changed" page from github. Looks like a cleanup.... I trust the rest is similar. 

kerolasa you broke the build on bsd or sunos a few patches back. Can you look into that? (Is it already fixed? revert that part? etc.)

kerolasa:
Let me see if I can find some bsd to try compilation. Sunos / solaris is more tricky, I cannot think anyone who would have one. Same goes for hpux, and so many olden times unixes.

PR #149, opened 2016-09-17
Mention + and - keys in the man page

jnweiger:
Found +. -. Ctrl-L keys and described their function under the --curses option.
Maybe also add these keys to the output of the onscreen help?

### DIFF
--- a/mtr.8.in
+++ b/mtr.8.in
@@ -190,6 +190,16 @@ Use this option to force
 .B mtr 
 to use the curses based terminal
 interface (if available).
+In case the list of hops exceeds the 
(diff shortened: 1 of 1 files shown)


rewolff:
Feel free to privide the patch for the onscreen help!
But... we can't tell EVERYTHING in the short onscreen help. So if we run out of space, we should just leave it. 
So: helping users by providing a few pointers to things they can do is good, but if there are a lot of things a user can do we should stop trying to mention everything. 
"you can hit A to insert the character A into the text" as onscreen help for a text editor is not the direction we want to go....

PR #151, opened 2016-09-29
2016-09-29 BSD build fix and minor improvements

kerolasa:
This should be straightforward to review.  Commit messages has further detail what is done and why.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -44,7 +44,9 @@ AC_CHECK_HEADERS([ \
   linux/icmp.h \
   ncurses.h \
   ncurses/curses.h \
+  netinet/in.h \
   socket.h \
+  sys/limits.h \
   sys/socket.h \
   stdio_ext.h \
   sys/types.h \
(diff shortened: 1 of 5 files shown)


rewolff:
I like patches that remove more code than they add. :-) 
(according to git ALMOST. In this case good enough!...)

kerolasa:
Thank you @grawity I did mess up the initialization, and here is updated version of @rewolff demo code that works the way earlier did.

    #include<stdio.h>
    #define NUM 10
    struct test {
    int x;
    int y[NUM];
    int z;
    }; 
    struct test bla = {
    .y = { [0 ... NUM - 1] =  -2}
    };
    int main (int argc, char **argv)
    {
     int i;
     printf ("bla.x = %d\n", bla.x);
     printf ("bla.y = ");
     for (i=0;i<NUM;i++) {
      printf ("%d ", bla.y[i]);
     }
     printf ("\nbla.z = %d\n", bla.z);
    }

I will send update about net_reset() code shortly.

PR #153, opened 2016-10-02
net: fix net_reopen() initialization

kerolasa:
Mantas Mikulenas reviewed recent pull reques that contained commit
b90a522f23167bd00062504803e94220937aba23 making a code path quicker.
Unfortunately initialization was incomplete, as demonstrated by R.E.
Wolff in same review.  Fix is to use GCC designated initializer.

Reference: https://github.com/traviscross/mtr/pull/151
Reference: https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html

### DIFF
--- a/net.c
+++ b/net.c
@@ -1492,16 +1492,17 @@ extern void net_reopen(struct mtr_ctl *ctl, struct hostent * addr)
 
 extern void net_reset(struct mtr_ctl *ctl)
 {
-  static const struct nethost template = {
-    .saved = { -2 },
-    .saved_seq_offset = 2 - SAVED_PINGS,
-    .xmit = 0
+  static struct nethost template = {
+    .saved_seq_offset = 2 - SAVED_PINGS
   };
 
-  int at;
+  int at, i;
 
   batch_at = ctl->fstTTL - 1;	/* above replacedByMin */
   numhosts = 10;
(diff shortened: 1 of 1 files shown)


rewolff:
Is this standard-C? Or only a GCC extension? Your comment points at the latter. 
Do we ONLY support gcc? I would not know off the top of my head a case that compiles with non-gcc, but I would prefer to keep compatibility with non-gcc compilers. 

So either you point me to a standards document that says this is standard C. 
Here (post#7) is a hint that it's a gcc extension: https://ubuntuforums.org/showthread.php?t=2225167

You might think it's a shame that standard C doesn't support this. And you're right. But it's a fact that we can't change. 

So... We have to drop the "const" qualifier, and upon startup loop to fill the array..... 

```
 // We'd like to initialize the whole array statically to the -2 value, but
 // alas, the syntax for that is an gcc extension. 
 .saved = { /* [0 ... SAVED_PINGS - 1] = */ -2 },

```

kerolasa:
Designated inits work with clang, but I guess supporting all sorts of exotic and/or old compilers is required.  I updated my pull request branch to use iterative loop, see commit 772afcec1f9d6c8d9790e0346d8b792df6735bb3

PR #157, opened 2016-10-08
Initialize dns process before opening display

### LINKED ISSUE #156: ncurses: Resizing terminal window freezes the program
This seems to be caused by the new dns implementation somehow (commit 02a9e6193e5dec01198f7824b7ed92f320b4e7a9).

Narthorn:
By opening the display before the dns process is forked, the child
inherits things like ncurses's handler for SIGWINCH (window resize
event), which can cause it to crash and leave the parent process in
an infinite select() loop, freezing the display.

This fixes #156.

### DIFF
--- a/mtr.c
+++ b/mtr.c
@@ -786,8 +786,8 @@ extern int main(int argc, char **argv)
 
 
     lock(stdout);
-      display_open(&ctl);
       dns_open(&ctl);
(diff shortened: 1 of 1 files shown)


rewolff:
Thanks for finding and fixing this!

PR #158, opened 2016-10-08
Add displaymode 2 back in

### LINKED ISSUE #155: --displaymode 2 not available in current HEAD version
The mtr man page claims 3 display modes via the --displaymode option:

>        --displaymode MODE
>               Use this option to select the initial display mode: 0 (default) selects statistics, 1 selects the stripchart without latency information, and 2 selects the stripchart with latency information.

However, trying to use --displaymode 2 with the current HEAD version of MTR results in an error:

> % mtr --version
> mtr 0.87.148-c96f
> % mtr 192.0.2.1 --displaymode 2
> mtr: value out of range (0 - 1): 2

Furthermore, running "mtr 192.0.2.1", then pressing "d" twice to cycle through displaymodes only lets me cycle through two modes: statistics mode, and stripchart without latency info.

In mtr 0.87, it was possible to use --displaymode 2, and it was possible to use "d" to cycle through 3 displaymodes as documented.

I haven't attempted to bisect, but did notice there's been some overall code cleanup in progress.  Wanted to make sure that exclusion of --displaymode 2 was not intentional.

Narthorn:
Issue introduced by 2e4369a61c5bf5a6c221a49350e71f8dfebba539, displaymode 2 was missed when converting from ints to an enum.

This fixes #155.

PS: This kind of breakage is not unexpected considering the ridiculous size of the pull request which introduced it... Looks like those commits should be reviewed more thoroughly.

### DIFF
--- a/display.h
+++ b/display.h
@@ -46,6 +46,7 @@ enum {
 enum {
   DisplayModeDefault,
   DisplayModeBlockmap,
+  DisplayModeBlockmapScale,
   DisplayModeMAX	/* this must be the last DisplayMode entry */
 };
 


rewolff:
Narthorn, I have little time to work on MTR myself. After the umptieth complaint about DNS problems I went along and rewrote it, the way I think it should have been done all along. You catched a bug in there. Thanks!. 
Now Sami was getting interested in contributing and he proposed a few patches. Some rewrites are inherently "big". And I missed this one in reviewing. And things that I do catch are sometimes deferred: "Patch accepted, but please fix .... ". That fixing then sometimes doesn't happen. Sorry. 

In any case, thanks for paying attention!

PR #159, opened 2016-10-11
2016-10-11 let the dust settle

kerolasa:
This pull request is full of cleanups and clarifications, some compiler warning fixes, and few option parsing fixes that I revisited thanks to Narthorn. Quite honestly I think next thing I should do is to stop sending pull request until a next release is out. Assuming of course bugs are not reported, leading to need to fix stuff that I have potentially broken.

### DIFF
--- a/curses.c
+++ b/curses.c
@@ -18,19 +18,19 @@
 
 #include "config.h"
 
+#include "mtr.h"
+
 #include <strings.h>
 #include <unistd.h>
 
 #include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 
 /* MacOSX may need this before socket.h...*/
 #if defined(HAVE_SYS_TYPES_H)
 # include <sys/types.h>
-#else
-/* If a system doesn't have sys/types.h, lets hope that time_t is an int */
-# define time_t int
 #endif
 
 #include <sys/socket.h>
@@ -64,7 +64,6 @@
 #include "display.h"
 #include "utils.h"
 
-#include <time.h>
 
 enum { NUM_FACTORS = 8 };
 static double factors[NUM_FACTORS];
--- a/curses.c
+++ b/curses.c
@@ -512,18 +512,17 @@ static void mtr_curses_init(void) {
 	block_map[NUM_FACTORS-1] = '>';
 }
 
-
-static int block_col[NUM_FACTORS+1] =
-{	/* 1:black 2:red 3:green 4:brown/yellow 5:blue 6:magenta 7:cyan 8:white */
-        COLOR_PAIR(2)|A_BOLD,
+enum { black = 1, red, green, yellow, blue, magenta, cyan, white };
+static const int block_col[NUM_FACTORS + 1] = {
+        COLOR_PAIR(red)|A_BOLD,
 	A_NORMAL,
-	COLOR_PAIR(3),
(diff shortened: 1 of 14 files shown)


curses.c
```diff
@@ -81,206 +97,214 @@ static void pwcenter(char *str)
...
-     return ActionClear;
-  if ((c==19) || (tolower (c) == 'p'))
-     return ActionPause;
-  if ((c==17) || (c == ' '))
-     return ActionResume;
-  if(tolower(c) == 'r')
+  case 12:
+    return ActionClear;
+  case 19:
+  case 'p':
```
rewolff: You need a case 'P': here too!

display.c
```diff
@@ -192,104 +190,40 @@ extern int display_keyaction(struct mtr_ctl *ctl)
...
+  if (ctl->DisplayMode == DisplayRaw)
     raw_rawhost (ctl, host, ip_addr);
-    break;
-  }
 }
 
 
 extern void display_loop(struct mtr_ctl *ctl)
 {
-  switch(ctl->DisplayMode) {
```
rewolff: The advantage of writing it like this is that it is similar in the different routines. All have a case like this and list the various calls for each of the display mode cases.

display.c
```diff
@@ -192,104 +190,40 @@ extern int display_keyaction(struct mtr_ctl *ctl)
...
-  case DisplayRaw:
+  if (ctl->DisplayMode == DisplayRaw)
     raw_rawhost (ctl, host, ip_addr);
-    break;
-  }
 }
 
 
 extern void display_loop(struct mtr_ctl *ctl)
 {
```
rewolff: The advantage here of course is that it's shorter.

curses.c
```diff
@@ -628,25 +627,11 @@ extern void mtr_curses_redraw(struct mtr_ctl *ctl)
 
   move(0, 0);
   attron(A_BOLD);
-  pwcenter("My traceroute  [v" PACKAGE_VERSION "]");
+  snprintf(buf, sizeof(buf), "%s%s%s", "My traceroute  [v", PACKAGE_VERSION, "]");
```
kerolasa: Earlier string was read-only, and pwcenter() needs write access. Yep, minor thing but necessary when all warnings were removed.

display.c
kerolasa: Fair comment.

display.c
kerolasa: In deed. And earlier code could have been cleaned up using default, but this is now as it is. If you want I can convert back to switch case + default. Downside is few lines more, and quite frankly this is just a style question and maintainers opinion is the law. Just tell if change is needed.

mtr.c
```diff
@@ -590,14 +590,13 @@ static void parse_arg (struct mtr_ctl *ctl, int argc, char **argv)
...
 {
-  int argc;
+  int argc = 1;
   char *argv[128], *p;
 
   if (!string) return;
-
-  argv[0] = "mtr";
+  argv[0] = xstrdup(PACKAGE_NAME);
   argc = 1;
```
kerolasa: I don't understand the question. This is again remedy to read-onlly data. If question is  static string from config.h rather than "mtr" then my answer is as simple as IMHO it is always best to avoid all magic strings and numbers, and prefer single symbolic definitions instead.

PR #162, opened 2016-10-27
Respect theme foreground color

vaygr:
This fixes the bug about mtr being almost unusable with dark GTK+ themes. Now non-down rows are colored based on whatever current theme offers instead of `black`.


rewolff:
Is it that simple??? :-)

PR #165, opened 2016-10-29
2016-10-29 usage improvements

kerolasa:
Make all compile time conditional options to behave same way, and add --sctp to manual page.


kerolasa:
Ping

rewolff:
pong

PR #166, opened 2016-11-02
Fix repeated DNS queries when toggling ipinfo

meingtsla:
Since 0.87, running mtr in interactive mode without an ipinfo option and then toggling an ipinfo display results in the same DNS TXT queries for each hop being sent on every cycle.  This is because no hash table is set up to store the results unless an ipinfo option is specified on startup.  To address this, always set up the hash table to prepare for the possibility of ipinfo being enabled.


meingtsla:
Figure it's time to bump this. :) I originally opened this prior to the code rebase work - so let me know if any changes are needed.

rewolff:
From the other pull request, I got the impression that it would always send out the queries. Just to confirm that this does not send out the queries when not needed. (If I don't want it on screen don't bother the DNS servers about it).

zorun:
Thanks @rewolff ! No, in any case no query is sent unless the ipinfo is displayed.

PR #171, opened 2016-12-25
split raw socket access into mtr-packet subprocess

matt-kimball:
With these changes, all raw socket access is split out from the main `mtr` process into a subprocess called `mtr-packet`.  There are a few advantages to these changes:

1.  Increased security.  The main `mtr` binary no longer needs to be SUID root, or to have any special capabilities.  This is nice, because it links into external libraries, such as ncurses and GTK+, which have large code-bases and potentially contain security flaws which might allow an attacker to get access to a raw network socket.  `mtr-packet` doesn't link with any such large libraries.  Also, the `mtr-packet` implementation now drops any special capabilities granted through the `setcap` commandline tool, which further hardens it.

2.  Ported to Windows (with Cygwin).  `mtr-packet` has two network implementations.  One is used for Unix-like operating systems, such as Linux, FreeBSD and MacOS.  This is similar to mtr's existing raw socket implementation.  There is now a second implementation for Windows, which uses Windows' `IcmpSendEcho` API to send network probes.  I believe this is the first time mtr has been working under Windows.

3. `mtr-packet` can be used to send many simultaneous network probes by programs other than `mtr`.  The method of using `mtr-packet` by other programs is documented with a man page.  It is my hope that this might lead to new and interesting network monitoring tools, with features and user interfaces which differ from `mtr`.

There is a lot of code changed here—several thousand lines of changed code.  I have put a lot of effort into keeping all the existing functionality of `mtr` intact, but with a change this size it is hard to avoid overlooking a detail or two.

I have tested on Linux, FreeBSD, MacOS and Windows.  I consider these to be the four most important operating systems to support in 2016.  It's likely that I broke things for, say, Irix from the mid-90's.  My apologies for anyone affected on operating systems other than those I tested on.


rewolff:
Matt, this is the direction I've been wanting MTR to go for years, but haven't been able to find the time program it myself. 
So I think it's important to have it more in the spotlight for some testing. 
I WOULD like to first do a "stable" release and THEN merge the "new features"/"code rework", but we still have a build regression on some OS that I don't have. IIRC that is macos, so with a bit of luck your work fixes that too. 

My proposal is to "just merge into master". Agreed?

matt-kimball:
Building on MacOS and FreeBSD with ncurses is still difficult due to change 4e2a948.  This pull request does not address that issue.

Just merging this into master seems fine to me.  But I would hold off on making a release with this until the dust has had some time to settle.

I can prepare a fix for the ncurses on MacOS issue some time this coming week.

yvs2014:
A little bit about library dust.
NetBSD, for example, is not using ncurses. But in systems of this type, you could use let's say BSD curses(3) by checking for the presence of such libraries.
Take look at lines 46-64 of https://github.com/yvs2014/mtr/blob/master/configure.ac

matt-kimball:
Sounds good to me.

Do you want to do the pull request?

matt-kimball:
Looking at the history, the configure.ac looked like your version before 31395df.

31395df changed it from that, seemingly to fix a build issue on Solaris.

Then 4e2a948 changed it again to depend on pkg-config.

Then I changed it again in f8e6dfa to not depend on pkg-config.

Maybe we want to keep the change from 31395df, but revert f8e6dfa and 4e2a948?  I don't know.

What I can tell you is that as of f8e6dfa, mtr builds and finds ncurses correctly on Ubuntu, FreeBSD, MacOS and Cygwin.  I was testing only on those four OSes.

yvs2014:
Some time ago I tested that on ubuntu, freebsd/netbsd and openindiana. Unfortunately, I'm not familiar with the logic of the current version.(

rewolff:
IMHO the configure logic is too complicated. I would prefer a rewrite to something alnog the lines: 
```
look for ncurses
if no ncurses found, look for curses. 
```
I haven't looked at it recently, but from what I remember there are a whole lot of nested functions that make it difficult for me to figure out what is going on. Apparently it worked for quite a while, and changing that stuff is dangerous. My fear of touching that code is apparently warranted.....

PR #172, opened 2016-12-26
build: use AC_CHECK_LIB for ncurses, rather than pkg-tool

matt-kimball:
pkg-tool is indeed an elegant way to check for the presence
of libraries.  When using pkg-config to check for ncurses, this
works great... for Linux.  Unfortunately, on FreeBSD and MacOS,
ncurses is installed by default, but pkg-config is not.  When
pkg-config is installed on these systems, it doesn't know about
the system installed ncurses.

Therefore, somewhat counterintuitively, it is better for portability
to just us AC_CHECK_LIB instead of PKG_CHECK_MODULES to find
ncurses.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,16 +95,7 @@ AC_ARG_WITH([ncurses],
   [AS_HELP_STRING([--without-ncurses], [Build without the ncurses interface])],
   [], [with_ncurses=yes])
 AS_IF([test "x$with_ncurses" = "xyes"],
-  [PKG_CHECK_MODULES([NCURSES], [ncurses], [
-      AC_DEFINE([HAVE_NCURSES], [1], [Define if ncurses library available])
-      AC_CHECK_LIB([ncurses],
-        [use_default_colors],
-        [AC_DEFINE([HAVE_USE_DEFAULT_COLORS], [1],
-          [Define this if your curses library has the use_default_colors() command.]
-        )
-      ])
-    ],
-    [with_ncurses=no])
+  [AC_CHECK_LIB([ncurses], [initscr], [], [with_ncurses=no])
 ])
 AM_CONDITIONAL([WITH_NCURSES], [test "x$with_ncurses" = xyes])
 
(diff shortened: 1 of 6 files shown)


rewolff:
Great! thanks.

yvs2014:
Unfortunatelly, building on freebsd without extra pkg-conf still doesn't work:
% ./bootstrap.sh
configure.ac:24: error: Could not locate the pkg-config autoconf macros.  These are usually located
in /usr/share/aclocal/pkg.m4.  If your macros are in a different location,
try setting the environment variable ACLOCAL_OPTS="-I/other/macro/dir"
before running ./bootstrap.sh again.
configure.ac:24: the top level
autom4te-2.69: /usr/local/bin/gm4 failed with exit status: 1
aclocal-1.15: error: echo failed with exit status: 1

And some another issue, mtr fails to start on systems without IPv6 support:
% mtr -4 8.8.8.8
Failure opening ICMPv6 send socket: Address family not supported by protocol family
mtr: Failure to start mtr-packet: Invalid argument
%

matt-kimball:
Thanks for trying this out.

Using `bootstrap.sh` will fail without pkg-conf installed.  I think if you already have a generated `configure` script, like someone would have with a tar of a full release, it will work without requiring `pkgconf`.  That was my intent, anyway.  Do we think it is important to be able to bootstrap without pkg-conf installed?

You probably already know, but you can get bootstrap to work if you run `pkg install pkgconf` on your FreeBSD system.

I'll work on the IPv6 issue.  I didn't think there were many systems without IPv6 support any more.  I tried to make things work if you don't have any IPv6 interfaces or routes.  But I didn't expect to to have a modern OS where you can't create an IPv6 socket.

yvs2014:
Yeah, maybe you're right, and it's just a matter of taste. If I recall correctly, in previous versions you could build it without pkg-config and use on systems without ipv6.

matt-kimball:
Maybe @kerolasa has an opinion on the value of using pkg-config, since I think he made the original change which added more dependenices on it.  This change removed that for ncurses, but kept the dependency on it for GTK.  

I don't have strong feelings, as long as I can build mtr with ncurses on FreeBSD and MacOS, without doing silly things like installing a second copy of ncurses so pkg-config can find it.  (Fixing that was what motivated this change.)

I will fix the IPv6 thing.

PR #173, opened 2016-12-27
man: simplify groff source for manual page synopsis

kerolasa:
Use .SY macro to describe synopsis, and tell about long-only options that
are far more readable than short options.

Signed-off-by: Sami Kerola <kerolasa@iki.fi>

### DIFF
--- a/man/mtr.8.in
+++ b/man/mtr.8.in
@@ -2,114 +2,47 @@
 .SH NAME
 mtr \- a network diagnostic tool
 .SH SYNOPSIS
-.B mtr
-[\c
-.BR \-4 |\c
-.B \-6\c
-]
-[\c
-.BI \-F \ FILENAME\c
-]
-[\c
-.B \-\-report\c
-]
-[\c
-.B \-\-report-wide\c
-]
-[\c
-.B \-\-xml\c
-]
-[\c
-.B \-\-gtk\c
-]
-[\c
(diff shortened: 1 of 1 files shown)


matt-kimball:
This does not work with groff 1.19.2, which is what MacOS still uses.

rewolff:
I was going to write: I bet this won't work on some older system. Matt already tested which one. 

You could write a "workaround" that converts the new ".OP" to what the old version used to use and we have an .8.src. Then for modern system the .8.src gets copied to .8, and for older systems we sed -e 's/NEWCONSTRUCT/OLDCONSTRUCT/" ...8.src > .8

kerolasa:
I didn't realize SY would have any portability issues. Oh well, if the macro unable adding a workaround to sounds wrong way to make manual more simple. Its bes to abandon this change.

matt-kimball:
Yeah, that is best.

I do think the goal of making the man pages more maintainable is a good one.  It would be nice if we could use some tool like ronn to write man pages in a readable format and then convert them to groff.  (See https://github.com/rtomayko/ronn)

However, using ronn in particular would put a build dependency on having Ruby installed, which I don't like.  Maybe there is some other tool build on something more likely to be installed everywhere (perhaps bash or python 2 or C source that we can include and build) that accomplishes something similar, but I don't know what tool that would be.

PR #177, opened 2017-01-03
Remove unnecessary MTR_PACKET env variable.

kasper93:
It is not nice to execute program based on some env variable. Also the
workaround for (non)issue make it harder to work with mtr without
installing in my opinion.

First we try mtr-packet from PATH. In case it is not found we try binary
from current directory. This way user can easily overwrite selected
binary by modifying PATH env variable. But we don't require having
either mtr or mtr-packet in PATH.

Also fix batch file to support paths with spaces on Windows.

------------------------------------

I'm open for discussion, but I find current design little bit clunky.

### DIFF
--- a/build-aux/mtr.bat
+++ b/build-aux/mtr.bat
@@ -18,15 +18,15 @@ rem  Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 rem
 
 rem Assume the path of this batch file is the mtr installation location
-set MTR_DIR=%~dp0
+set "MTR_DIR=%~dp0"
 
-set MTR_BIN=%MTR_DIR%\bin
+set "MTR_BIN=%MTR_DIR%\bin"
 
 rem ncurses needs to locate the cygwin terminfo file
-set TERMINFO=%MTR_DIR%\terminfo
+set "TERMINFO=%MTR_DIR%\terminfo"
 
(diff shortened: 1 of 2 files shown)


matt-kimball:
I was thinking of `MTR_PACKET` as an analogue to the `VISUAL` or `PAGER` environment variables.  By allowing a full filename name to be specified, you could use something which doesn't necessarily have the filename "mtr-packet".

This would allow using some other tool with the same interface as mtr-packet, but with a different implementation.  Perhaps that is overly optimistic about the existence of some future tool like that.

Using `execlp` instead of `/usr/bin/env` is a good change.

rewolff:
Oh, that would allow me to test in my build directory by setting it to ./mtr-packet, instead of` ln -s /home/.... build-directory/mtr-packet /usr/bin ` .

kasper93:
> I was thinking of MTR_PACKET as an analogue to the VISUAL or PAGER environment variables. By allowing a full filename name to be specified, you could use something which doesn't necessarily have the filename "mtr-packet".

Fine, let's keep `MTR_PACKET` variable. I still changed to `execlp` and if this one fails try to run executable from current directory not only if `/usr/bin/env` is not available like it was. I'm simple man I want to do `make && ./mtr -tr github.com` without required intermediate steps to setup env :)

PR #180, opened 2017-01-06
couple maintenance changes

kerolasa:
All pretty trivial. Should be pretty straightforward to review.


matt-kimball:
Looks good to me.

I always build in a separate directory from the source directory, so I didn't notice those .gitignore omissions.  Good clean-up.

PR #188, opened 2017-01-27
Improve readability of net.c

jbruggeman:


rewolff:
The original code also had spaced around ALL braces of the IF statements. I've added a bunch of those. 
And in my book you cannot trust the compiler when it comes to evaluating if ( a < b || c < d) : some compilers get the evaluation order wrong (or maybe it was the C-standard?) so you'd get the wrong result. I've learnt to add parentheses, just to be sure.

PR #215, opened 2017-08-13
mtr-packet: use ICMP and UDP without privilege on linux

linghutiantian:
This commit enables non-privileged users to use mtr on linux without
setuid. Currently ICMP and UDP protocols are supported in this commit.
Previously, to use mtr on linux with protocol ICMP and UDP, RAW sockets
have to be opened to send out packets and receive ICMP errors, so users
must have RAW socket permission to use this program. The goal of this
commit is to make mtr usable for normal users without RAW socket
permission. A sysctl might need to be set to make this work,
https://lkml.org/lkml/2011/5/18/305
The changes include:
(1) The origianl logic is not changed, but instead, when the program
fails to open RAW sockets, it will fallback to opening DGRAM scoekts.
(2) A new flag is created to indicate whether RAW socket is used for
IPv4 and IPv6 respectively.
(3) When using DGRAM sockets to send out packets, receive sockets are
not required. Instead, IP_RECVERR is enabled to receive ICMP errors.
(4) Packet receiving function is changed from recvfrom() to recvmsg() to
retrieve more information.
(5) When error is indicated, the program will check the error type and
read from the error queue of the socket. Original payload causing the
error will be read out from the error queue to match for probe, and
additional data (e.g. source ip of ICMP error packets) will be retrieved
from CMSG.
(6) Use a separate socket to probe byter order if raw socket creation
fails, to avoid double bind issue.
(7) Also a few tweaks are added to make non-RAW socket working.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -46,6 +46,7 @@ AC_CHECK_HEADERS([ \
   error.h \
   fcntl.h \
   linux/icmp.h \
+  linux/errqueue.h \
   ncurses.h \
   ncurses/curses.h \
   netinet/in.h \
--- a/packet/command.c
+++ b/packet/command.c
@@ -321,6 +321,7 @@ void send_probe_command(
     param.ttl = 255;
     param.packet_size = 64;
     param.timeout = 10;
+    param.is_probing_byte_order = false;
 
     for (i = 0; i < command->argument_count; i++) {
         name = command->argument_name[i];
--- a/packet/construct_unix.c
+++ b/packet/construct_unix.c
@@ -138,8 +138,13 @@ void construct_icmp4_header(
     struct ICMPHeader *icmp;
     int icmp_size;
 
-    icmp = (struct ICMPHeader *) &packet_buffer[sizeof(struct IPHeader)];
-    icmp_size = packet_size - sizeof(struct IPHeader);
+    if (net_state->platform.ip4_socket_raw) {
+        icmp = (struct ICMPHeader *) &packet_buffer[sizeof(struct IPHeader)];
+        icmp_size = packet_size - sizeof(struct IPHeader);
+    } else {
+        icmp = (struct ICMPHeader *) &packet_buffer[0];
+        icmp_size = packet_size;
+    }
 
     memset(icmp, 0, sizeof(struct ICMPHeader));
 
@@ -224,8 +229,13 @@ void construct_udp4_header(
     struct UDPHeader *udp;
     int udp_size;
 
-    udp = (struct UDPHeader *) &packet_buffer[sizeof(struct IPHeader)];
-    udp_size = packet_size - sizeof(struct IPHeader);
+    if (net_state->platform.ip4_socket_raw) {
+        udp = (struct UDPHeader *) &packet_buffer[sizeof(struct IPHeader)];
+        udp_size = packet_size - sizeof(struct IPHeader);
+    } else {
+        udp = (struct UDPHeader *) &packet_buffer[0];
+        udp_size = packet_size;
+    }
 
     memset(udp, 0, sizeof(struct UDPHeader));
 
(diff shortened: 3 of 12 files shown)


rewolff:
The original Unix-authors knew very well that "sending ICMP packets" is a very dangerous tool. 

The "normal" uses of that privilege are quite boring: Sending packets that test the functionality (connectivity) of the network. 

But in the wrong hands, this privilege is quite a lot stronger. 

MTR uses the privilege to create (slightly broken) packets that the normal IP stack would never generate. This is used to simply test the network. But there is no check for "slightly broken" versus "completely broken". The original Unix-authors decided: Let's leave the responsibility  between "correct", "slightly" and "completely broken" to the user space programs like ping and mtr. 

So... what can you do with an ICMP socket? For one you can send an "ICMP connection refused". Possibly with forged from address. This will tear down an existing TCP connection, possibly between two unrelated hosts. 

In the early 1990-ies. the TUDELFT university network was internally unrouted. A hardware bug in a PC a few floors up would cause that PC there to receive a packet that was not intended for it. The software neglected to check the destination address (assuming the hardware should've done that) and reported: ICMP: I have no knowledge of this connection you are talking about. This would tear down a completely valid connection between two hosts that had nothing to do with that PC.... 

So... A hacker might use the privilege to send ICMP packets to flood your PC with "ICMP: you are NOT talking to www.amazon.com". 

Mark my words: In a few decades there will be a big stink about this.

linghutiantian:
Totally agree. However if a user has to use mtr, giving him "icmp socket privilege" is still better than giving him full "raw socket privilege" :)

rewolff:
Review required... (by me).

rewolff:
Will this create problems on other platforms like macos? Windows? Matt... are you able to test this on either?

aquerubin:
Mtr runs on Linux without setuid already.  Did something change?

Tony

rewolff:
Unix restricts certain operations to "root only" . This is because they would compromise system- or network-security. 
Now simple things like "read a file" could  be considered like that as well. But the kernel has builtin logic: if you're allowed to read the file, go ahead.

For stuff like "raw sockets" the kernel writers realize that writing the filters to differentiate between "harmless" and "harmful" is difficult. So... they decided to make the whole "raw sockets" root-only. 

Now a program like "ping" needs to be setuid to work. That allows it to open a raw socket and do its thing. Now the ping executable is the "guardian" that is supposed to guard between harmful and harmless uses of the privileges. The system as  a whole trusts the PING executable not to allow malicious use of the "raw socket" that it has access to. 
Similarly, mtr is setuid-root or set-capability-raw-socket-access. 
With that new kernel feature we'd be able to make that setgid-to-the-group-that-has-raw-socket-access. 

From mtr's point of view that is a lowering of security risks. But for the kernel / linux system as a whole, the "fact" that a few years from now we'll see a distribution that distributes say the apache user as group netsocket, and then remote users who hack the webserver suddenly have the ability to open raw network sockets..... 

I'll buy you a beer in 20 years if there hasn't been a security breach due to these developments. I should say 30 years to have a bigger chance of winning the bet, but the chances of me being around for you to collect  are also dropping by that time...  :-)

matt-kimball:
I approve of the goal of this change.

However, with this change, mtr fails to build on both Mac OS and Windows.  On Mac OS, `SOL_IP` is undefined.  On Windows, `ip4_socket_raw` and `ip6_socket_raw` are not members of `net_state->platform`.

At a minimum, a large change such as this should be working on Mac OS, Windows, FreeBSD and Linux, before merging.

linghutiantian:
Thanks for pointing this out. I will fix the Windows error. But I don't know how to solve the undefined SOL_IP problem on Mac OS, can we selectively compile this code out just for MacOS and force it to use raw socket only?

rewolff:
Supposedly autoconf would allow you to test at configure time if SOL_IP is available. 

Make your code depend on "HAVE_SOL_IP", make sure it compiles WITH and WITHOUT this.

If you don't know how to make autoconf automatically provide this define when it's available... that makes two of us. I don't know either.

adpoliak:
> can we selectively compile this code out just for MacOS and force it to use raw socket only

> If you don't know how to make autoconf automatically provide this define when it's available

My [PR](https://github.com/traviscross/mtr/pull/216) does this to bypass UID 0 checks when building under Cygwin.  
The `AM_CONDITIONAL()`, `AM_COND_IF()`, and `AC_DEFINE()` constructs would be used if the condition it can be tested via the shell.  
The `AC_CHECK_FUNC()`, `AC_CHECK_LIB()`, and `AC_DEFINE()` constructs would be used if the condition has dependencies in libraries/include files that may/may not be present in the build system.

linghutiantian:
Thank you all for your help. I have fixed the compilation issues for both OS X and Windows, and updated the pull request. Please let me know what you think about this PR. Thanks.

PR #216, opened 2017-08-16
add Cygwin build conditional to skip UID 0 checks

adpoliak:
Uses the C preprocessor to skip code that performs checks against UID 0 using `getuid()` when building under Cygwin.

Tested working against Cygwin x86_64 (no UID 0 checks are performed) and macOS Sierra (macports toolchain, UID 0 checks are performed).

GTK is untested, but should work.


rewolff:
I implemented this differently. Now there is only one place where the getuid is conditional.

PR #218, opened 2017-08-26
Update incorrect FSF address

robert-scheck:
The address of the Free Software Foundation mentioned in various files of the mtr source code is no longer correct. The new address is:

> Free Software Foundation
> 51 Franklin Street, Fifth Floor
> Boston, MA 02110-1301
> USA

Source: http://www.fsf.org/about/contact/

Source files, that previously used the layout of the default code header are doing so further, even the formatting was changed by the FSF due to the address change. Source files which used a custom layout for the code header are doing so further.


rewolff:
thanks for doing the legwork!

PR #222, opened 2017-09-06
mtr-packet: fix a bug causing IPv6 raw socket not working

linghutiantian:
This bug causes mtr not working on IPv6 only machines if invoked by root
user, and it is fixed in this commit.


rewolff:
Good spotting! Thanks!

PR #228, opened 2017-10-15
[typo]mtr.8.in: s/allows to/allows one to/

samueloph:
This fixes a very small typo discovered by lintian while i was packaging 0.92 for debian


rewolff:
It depends a bit on your english accent if this is a typo or not.

samueloph:
@rewolff, yeah, it looks like the kind of things that are on the gray area, i thought it didn't hurt to change to "allows one to" either way to make sure.

Thanks for accepting the change.

PR #248, opened 2018-03-21
Update mtr-packet.8.in

SaintBol:
RFC 5462 - MPLS EXP Field renamed to TC Field (Traffic Class)


rewolff:
Hi, Thanks for bringing things up-to-date with current standards. But next time, these should go in one pull commit, and therefore in one pull request.

PR #255, opened 2018-04-23
asynchronous resolve ipinfo 

weiyixuan:
Change the access to ipinfo to asynchronous mode.
It has been tested on MAC, Linux, and MAC platforms.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -137,20 +137,15 @@ AC_CHECK_FUNC([socket], [],
 
 AC_CHECK_FUNC([gethostbyname], [],
   [AC_CHECK_LIB([nsl], [gethostbyname], [], [AC_MSG_ERROR([No nameservice library found])])])
+	   
+AC_CHECK_FUNC([pthread_create], [],
+  [AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([No pthread library found])])])
 
 # Find resolver library.
-AC_CHECK_FUNC([res_query], [RESOLV_LIBS=""], [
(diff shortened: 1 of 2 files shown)


rewolff:
Hi, 
I have bad experiences with "removing search-for-library" from the configuration stuff. It is there for a reason. So you might have tested it on an "uptodate" Linux distribution, but what if someone is running a 5 year old distribution? (I have been contracted to renew a system that was installed in 1996, updated in 2001. It is still running the release from june 2001... Some old stuff just stays around..... I just checked. The actual install happened on sept 20th 2001)

rewolff:
I am no longer sure how the normal resolve stuff works. 

I have been wanting to make that asynchronous for a long time: fork a process, do the lookup, report result to parent. Simply use the system library to do the lookups. I don't remember people complaining about IPV6 lookups not working for a while, so I think it's gotten implemented. 

So... Is -lresolv no longer necessary? did we remove that dependency? What if someone doesn't have -lcares? Is there a fallback to the older -lresolv?

weiyixuan:
I've roughly read the way the res_query implemented in glibc-2.27: encapsulating and sending DNS requests, and waiting for a response.
res_query - > res_mkquery/res_send - > blocking call socket()/writev()/read()/sendmsg()/recvfrom(), etc.

Did you mean to use -lresolv in the child process to simulate the implementation asynchronously?

weiyixuan:
Another question: in the DNS-child-process, a process is forked for each IP address resolution request? Why not queue up and deal with it sequentially?

PR #256, opened 2018-04-30
Modify the access method for ipinfo

weiyixuan:
Use child process to get the ipinfo and pass it to parent process, using the -lresolv library.

### DIFF
--- a/ui/asn.c
+++ b/ui/asn.c
@@ -21,6 +21,7 @@
 #include <unistd.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <fcntl.h>
 #include <sys/types.h>
 #ifdef HAVE_ERROR_H
 #include <error.h>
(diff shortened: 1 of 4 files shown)


ui/asn.c
```diff
@@ -313,10 +404,67 @@ int is_printii(
...
+    } else if (pid == 0) {
+        int i;
+        FILE *infp = NULL;
+        char buf[64];
+
+        if (signal(SIGCHLD, SIG_IGN) == SIG_ERR) {
+            error(EXIT_FAILURE, errno, "signal");
+        }
+
+        for (i = 3; i < fromres[1]; i++) {
```
rewolff: wouldn't it be better to start i at i=0 ?

rewolff:
I have added a comment on one line: Wouldn't it be better to just close fd 0-2 as well?

PR #259, opened 2018-05-07
for tcp, fix : bind port failed, try next sequence

weiyixuan:


rewolff:
I have rewritten, but not tested your patch. Can you test it for me?

weiyixuan:
The code can't be compiled, and the try_times needs to be modified to trytimes.
No problems were found after the test （rewritten code).

PR #261, opened 2018-06-05
Also try SOCK_RAW/IPPROTO_ICMP when other fail.

### LINKED ISSUE #223: mtr using IPPROTO_ICMP with SOCK_DGRAW instead of SOCK_RAW
Linux manual says that IPPROTO_ICMP should be used with SOCK_RAW:
" A user protocol may receive ICMP packets for all local sockets by
       opening a raw socket with the protocol IPPROTO_ICMP.  See raw(7) for
       more information. "

while mtr uses socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP);

Why? That doesn't work here under linux-vserver (socket() fails with "Permission denied" errno).

For now using
````
diff -urN mtr-0.92.org/packet/probe_unix.c mtr-0.92/packet/probe_unix.c
--- mtr-0.92.org/packet/probe_unix.c    2017-09-23 13:55:17.161190504 +0200
+++ mtr-0.92/packet/probe_unix.c        2017-09-23 13:53:07.961076290 +0200
@@ -276,7 +276,10 @@

     icmp_socket = socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP);
     if (icmp_socket == -1) {
-        return -1;
+        icmp_socket = socket(AF_INET, SOCK_RAW, IPPROTO_ICMP);
+        if (icmp_socket == -1) {
+            return -1;
+        }
     }
 #ifdef HAVE_LINUX_ERRQUEUE_H
     if (setsockopt(icmp_socket, SOL_IP, IP_RECVERR, &val, sizeof(val)) < 0) {
@@ -299,9 +302,12 @@

     icmp_tmp_socket = socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP);
     if (icmp_tmp_socket == -1) {
-        close(icmp_socket);
-        close(udp_socket);
-        return -1;
+        icmp_tmp_socket = socket(AF_INET, SOCK_RAW, IPPROTO_ICMP);
+        if (icmp_tmp_socket == -1) {
+            close(icmp_socket);
+            close(udp_socket);
+            return -1;
+        }
     }

     net_state->platform.ip4_present = true;
````

arekm:
Under Linux Vserver SOCK_RAW is quite limited. Only IPPROTO_ICMP works,
so try to use it.

check_length_order() also doesn't work because sendto is also limited.
Fortunately under linux check_length_order is not needed as described
in comment around check_length_order.

Fixes #223 for me.

### DIFF
--- a/packet/probe_unix.c
+++ b/packet/probe_unix.c
@@ -128,6 +128,12 @@ void check_length_order(
     ssize_t bytes_sent;
     int packet_size;
 
+#ifdef __linux__
+    /*  Linux will accept either byte order and check below fails to work
+     *  in some cases due to sendto() returning EPERM. */
+    return;
+#endif
+
     memset(&param, 0, sizeof(struct probe_param_t));
     param.ip_version = 4;
     param.protocol = IPPROTO_ICMP;
@@ -230,7 +236,10 @@ int open_ip4_sockets_raw(
 
     send_socket = socket(AF_INET, SOCK_RAW, IPPROTO_RAW);
     if (send_socket == -1) {
-        return -1;
+        send_socket = socket(AF_INET, SOCK_RAW, IPPROTO_ICMP);
+        if (send_socket == -1) {
+            return -1;
+        }
     }
 
     /*


rewolff:
I suspect "it used to work". So either it's a bug, or other OSes required the other way of doing things. 

I approve of "conservative" approaches where you leave the existing behaviour if it happens to work....

PR #269, opened 2018-09-23
Masking entries on the IPs/Hostnames list (curses interface)

appblue:
Hi,
Recently I needed a functionality, to mask initial part of the ip addresses/hostnames from the traceroute information, to hide local part of the network route and get something similar to this output, showing the result of the command:

` $ mtr --tcp -P 80 --curses github.com`

The output is:

```                             My traceroute  [v0.92.69-ff32]
hostname masked (ip masked)                                    2018-09-23T20:20:28+0200
Keys:  Help   Display mode   Restart statistics   Order of fields   quit

                             Last  58 pings
 1. ___ip_masked___                                                ....................
 2. ___ip_masked___                                                ....................
 3. ___ip_masked___                                                ....................
 4. ___ip_masked___                                                ....................
 5. ___ip_masked___                                                ....................
 6. ___ip_masked___                                                ....................
 7. pzn-b1-link.telia.net                                          .....1...1.......1.1
 8. hbg-bb4-link.telia.net                                         .1.......1.1........
 9. ldn-bb4-link.telia.net                                         11..11...1...111.11.
10. ash-bb4-link.telia.net                                         ............1.1.....
11. rest-b1-link.telia.net                                         ....................
12. github-ic-325491-ash-b1.                                       11.....111.....11..?
13. github-ic-325490-ash-b1.                                       ??1?.?>?1>??1..1???
14. (waiting for reply)                                            ???????????????????
15. lb-192-30-253-113-iad.gi                                       .111>1111>1?1???111

Scale:  .:117 ms  1:461 ms  2:1034 ms  3:1837 ms  a:2869 ms  b:4130 ms  c:5621 ms  >
```
I have added new interactive command '0' (zero), that allows to define how many of existing entries in the host names should be "masked", starting from the top.

Regards,
Chris

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ confinc
 confmf
 conftest.*
 stamp-h1*
+tags
+cscope.out
 
 /build-aux/compile
 /build-aux/depcomp
--- a/ui/curses.c
+++ b/ui/curses.c
@@ -279,6 +279,25 @@ int mtr_curses_keyaction(
         ctl->maxTTL = i;
 
         return ActionNone;
+    case '0':
+        mvprintw(2, 0, "Mask : %d\n\n", ctl->fstMASK);
+        move(2, 7);
+        refresh();
+        while ((c = getch()) != '\n' && i < MAXFLD) {
+            attron(A_BOLD);
+            printw("%c", c);
+            attroff(A_BOLD);
+            refresh();
+            buf[i++] = c;       /* need more checking on 'c' */
+        }
+        buf[i] = '\0';
+        i = atoi(buf);
+
+        if (i > (MaxHost - 1))
+            return ActionNone;
+        ctl->fstMASK = i;
+
+        return ActionNone;
         /* fields to display & their ordering */
     case 'o':
         mvprintw(2, 0, "Fields: %s\n\n", ctl->fld_active);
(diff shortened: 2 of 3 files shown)


yvs2014:
There's an option for skipping the initial part of mtr's output. What about the 'f' option?

appblue:
Hi @yvs2014! 

Thanks for your comment.

I completely agree that 'f' command is an option, but in my scenario I almost always want to share the complete trace route (with the response time histogram from the initial part of the output), but without exposing the IPs or hostnames of the local part of the network - I think it's not an uncommon scenario.

Regards
Chris

rewolff:
About "how to do this".... 

from all display modules, they HAVE an IP address (IPV6?) and need a string-to-display. So... if it isn't already like this all those need to call a "convert this to string in current settings". -n specified for numeric output? Then return a string like "8.8.8.8". No -n? "dns.google.com", masked? "--ip masked--". 

That would mean that after a bit of refactoring, you'd get ALL output modules in one go. 

So.... This is how I WANT it to be done..... Now I'll start looking at your patch.

appblue:
Thanks for your feedback. I’ll work on the change this week ;)

// ******************
Best Regards,
Krzysztof Kielak

rewolff:
OK. I've looked at your patch. It is specific for curses output. 
The mask is ONLY on the "near" end of the route. What if you want to mask off the other end?
How about: For internal use: the mask is a 64-bit bitmask. So for now you've implemented stuff at the near end: mask = (1 << i) -1; Sets the first "i" bits. Leave for later to implement: mask = -1 - (1 << i) .... (some complicated bitmagic I don't want to figure out completely right now) to implement "mask off further than i". 

Now the CORE can handle arbitrary masks, but we only support setting them in a few (one at first) basic ways. 

I understand that YOU are interested in the "curses" variant. But for now I would appreciate it if you could write it in such a way that it will be easy to make it  work for other output modules as well. It is acceptable if you only implement "dynamic changing of the mask" in the curses module. It is acceptable if you only modify the curses output module to use the new "convert to string" module. (but preferred if you'd do it for most or all). 

Can you do it for me like this?

appblue:
Hi,
I like the idea of the bitmap and I'll be more than happy to work on this to encapsulate this in a "module", so it could be used in all display modes and reports.

rewolff:
"module" is a bit a "too big" word. It's going to be more-or-less one function. On the other hand, with the dynamic DNS lookups that we're doing, things may be a bit more complicated than I thought. (The mapping IP addr -> name may change, also by the user chaning the mask..... ).

Issue #270, opened 2018-09-28
the freenode cygwin community strongly discourage using apt-cyg


divinity76:
had the following conversation on irc://irc.freenode.net/#cygwin , the tl;dr is 

> do not use apt-cyg - it's a horrible attempt at circumventing the package manager - it doesn't do dependency checks - it uses actual cygwin resources (need to upgrade mintty? bash? wget? GOOD LUCK!) - meaning rebasing fails - and has been known to irreversibly break installs. don't use apt-cyg or cyg-apt.

the full conversation conversation was a little closer to:

```irc
<hanshenrik> is mtr in a cygwin package?
<Redfoxmoon> hanshenrik, won't work; you can take a look at winmtr though.
<hanshenrik> seems it's not in cygwin, but the official github repo explains how to build it on cygwin, should be close enough ^^ 
<Redfoxmoon> oh goodness, they've ported it to cygwin
<hanshenrik> ("BUILDING FOR WINDOWS
<hanshenrik> " here https://github.com/traviscross/mtr )
<Redfoxmoon> oh dear.
<hanshenrik> ?
<Redfoxmoon> do not use apt-cyg
<hanshenrik> why? setup.exe is superior? :p
<Redfoxmoon> quite. apt-cyg will break your cygwin installation
<hanshenrik_> guess it's a sort of half-assed reimplementation of setup.exe then? don't see why they'd bother, given that it could easily have been a wrapper around setup.exe...
<furrymcgee> easily?
<MetaNova> hanshenrik_: it's not a half-assed reimplementation at all
<MetaNova> it's a horrible attempt at circumventing the package manager
<MetaNova> it doesn't do dependency checks
<MetaNova> it uses actual cygwin resources (need to upgrade mintty? bash? wget? GOOD LUCK!)
<MetaNova> meaning rebasing fails
<MetaNova> and has been known to irreversibly break installs. don't use apt-cyg or cyg-apt.
<hanshenrik_> furrymcgee, yeah, `apt-cyg install automake pkg-config make gcc-core libncurses-devel` could easily have been a wrapper around `setup.exe --wait --quiet-mode --packages  automake pkg-config make gcc-core libncurses-devel`  
<hanshenrik_> but by what you're saying here, that's obviously not the implementation they went with
<hanshenrik_> err, that list is supposed to be comma-separated for setup.exe, not space-separated
<hanshenrik_> anyway, then it would not circumvent the package manger, would handle dependencies, and probably wouldn't break anything
[...]

PR #271, opened 2018-09-28
use setup.exe's package manager mode, replacing apt-cyg

divinity76:
see #270 for rationale


rewolff:
I have no windows machines, no opinion on if this is desirable or not, but I'm taking your word on this....

PR #272, opened 2018-09-28
Update README

magnum7385:
Updated the cygwin setup executable name to its current 32bit version, setup-x86.exe, instead of setup.exe


divinity76:
(ick, sorry, in my defense, the name *used* to be setup.exe in the past)

rewolff:
That counts as confirmation.

PR #274, opened 2018-10-29
udp checksum is incorrect

commonism:
in certain circumstances -I'm aware of specifying localport and port in udp mode- mtr-packet sends udp packets with invalid checksum.

```
"33000 send-probe ip-4 1.1.1.1 local-ip-4 1.2.3.4 protocol udp size 64 bit-pattern 0 tos 0 ttl 1 timeout 10 port 123 local-port 33333 bit-pattern 111" | sudo ./mtr-packet
```

The changes correct this behaviour, they calculate the udp checksum and write to the required offset.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -98,7 +98,8 @@ mtr_packet_SOURCES = \
 	packet/probe.c packet/probe.h \
 	packet/protocols.h \
 	packet/timeval.c packet/timeval.h \
-	packet/wait.h
+	packet/wait.h \
+	packet/sockaddr.c packet/sockaddr.h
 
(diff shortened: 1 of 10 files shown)


commonism:
I updated the branch with more changes to improve mtr in udp mode.

commonism:
As these changes improve the udp mode by comparing more values when looking up a probe match, preventing invalid probe pick ups due to sequence number collisions when running multiple mtr in parallel, I'd be glad if they could get merged.
In case there are things to address first, please let me know.

rewolff:
Feel free to "prod" me earlier than after 9 months if I seem to have overlooked a pull request.

PR #281, opened 2018-12-06
Use monotonic clock.

dajohi:
This avoids a jumping system clock causing issues.

### DIFF
--- a/packet/timeval.c
+++ b/packet/timeval.c
@@ -16,7 +16,29 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
+#ifdef HAVE_CLOCK_GETTIME
+#include <time.h>
+#endif
+
 #include "timeval.h"
+
+/*
+    Return timeval using monotonic clock.
+*/
+int getmonotime(
+    struct timeval *tv)
+{
+#ifdef HAVE_CLOCK_GETTIME
+    struct timespec ts;
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0) {
+        return -1;
+    }
+    TIMESPEC_TO_TIMEVAL(tv, &ts);
+    return 0;
+#else
+    return gettimeofday(tv, NULL);
+#endif
+}
 
 /*
     Ensure that a timevalue has a microsecond value in the range
--- a/ui/select.c
+++ b/ui/select.c
@@ -62,9 +62,15 @@ void select_loop(
     struct timeval intervaltime;
     static double dnsinterval = 0;
 
-    memset(&startgrace, 0, sizeof(startgrace));
-
+#ifdef HAVE_CLOCK_GETTIME
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    TIMESPEC_TO_TIMEVAL(&lasttime, &ts);
+#else
     gettimeofday(&lasttime, NULL);
+#endif
+
+    memset(&startgrace, 0, sizeof(startgrace));
 
     while (1) {
         dt = calc_deltatime(ctl->WaitTime);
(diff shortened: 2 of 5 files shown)


rewolff:
I seriously doubt that all systems that we support will have the function that you use to get the monotonic clock. So, you've done things nicely that there is now one "getmonotime" function whose implementation could be what you've written, OR just "gettimeofday" when clock_gettime or TIMESPEC_TO_TIMEVAL does not exist. 

Ideally a compile time check would determine if clock_gettime is available and use the "better" interface when available.

If you can't hack autoconf to make it that way I need proof that this causes problems in practice before I commit such a change that is sure to cause problems on systems that don't have clock_gettime in the way that you expect....

ui/select.c
```diff
@@ -62,9 +62,15 @@ void select_loop(
...
 
-    memset(&startgrace, 0, sizeof(startgrace));
-
+#ifdef HAVE_CLOCK_GETTIME
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    TIMESPEC_TO_TIMEVAL(&lasttime, &ts);
+#else
     gettimeofday(&lasttime, NULL);
+#endif
```
krisl: can we use `getmonotime` here?

ui/select.c
```diff
@@ -123,8 +129,12 @@ void select_loop(
                 if (ctl->Interactive)
                     display_redraw(ctl);
 
+#ifdef HAVE_CLOCK_GETTIME
+                clock_gettime(CLOCK_MONOTONIC, &ts);
+                TIMESPEC_TO_TIMEVAL(&thistime, &ts);
+#else
                 gettimeofday(&thistime, NULL);
-
+#endif
```
krisl: can we use `getmonotime` here?

rewolff:
Look, I don't understand what the problem is that this solves. The systemtime is normally monotonic, it doesn't jump around when it wants to. Or say, "well, on tuesdays it is often really bad". No, time just goes forward all the time. 
Now when DOES the system time jump? Well if you go on vacation, turn off your computer for a week and your bios clock is really slow, so it has lost say 5 seconds. You start her up and almost immediately start an mtr to a host with one of the first hops having a 10 second delay. 

So while this is running, ntp finally trusts the measurements enough to jump the system time. from 10:05:23 time suddenly jumps to 10:05:28. Now packets sent at 10:05:21, suddenly arrive only at 10:05:36. getmonotime does nothing to correct this. 

If ntp jumps the time the other way, there might be some "faster" packets that get a corrected, but still wrong timestamp when they get back. On the long packets, you simply get wrong measurements again. If mtr does see and display these absurd values, the user can clearly see that there are measurement errors. Suppose this jump happens when I'm not looking. When I get back, I see a nicely progressing "minimum time" going from 0 to 1 second. That minimum time is what I often use to guess the "no congestion" delay of a path. When there are negative numbers in the list I know there is a measurement error. If all looks seemingly OK, I might be convinced that those numbers are real. 

getmonotime is meant for situations where time going backwards is really bad, Or when two consecutive timestamps should in no circumstance be the same, because the order in which things happened is very important for you. 

So..... I'm afraid such a change will break compatibility with some existing platforms that we support. Secondly I don't see the problem that it is supposed to fix.

packet/timeval.c
```diff
@@ -16,8 +16,30 @@
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
+#ifdef HAVE_CLOCK_GETTIME
+#include <time.h>
+#endif
+
```
levesquejf, line 22: I have to add `#include <stddef.h>` otherwise `NULL` is undefined.

rewolff:
For those wanting to look into this in the future. 

My software-engineering standpoint is that when it used to be "gettimeofday", and you improve that in some way on some systems when the improved stuff is available, then, instead of 
```
#ifdef SOMETHING 
   improved_stuff
#else  // github make typing that #else really difficult. 
   gettimeofday ();
#endif         
```
you should use
```
   GETTIMEOFDAY ()
```
inline in the running code, and somewhere in a header IF the right conditions exist, you can #define that into better_gettimeofday () or stuff like that. But "inline", the code is not being complicated beyond: "hey, why is that in capitals instead of the normal lowercase". If it is important, you can look up the definition of the GETTIMOFDAY macro, if it is not, you are not confronted with "compile time flow control" directives inside the code you are trying to understand.

PR #289, opened 2019-02-13
Fix disappearing URLs in man pages for MacOS

matt-kimball:
groff under MacOS is missing support for .UR and .UE codes, which are
used to embed URLs in man pages.  In fact, using those codes causes
the URLs to disappear from the man page entirely.

With this change, when we generate man pages, strip of the .UR and .UE
codes when building under MacOS.

### DIFF
--- a/build-aux/mangen.sh
+++ b/build-aux/mangen.sh
@@ -11,4 +11,19 @@ if [ $# -lt 3 ]; then
     exit 1
 fi
 
-sed -e "s|@VERSION[@]|$1|g" $2 >$3
+TMP_FILE=/tmp/mangen.$$
+
+sed -e "s|@VERSION[@]|$1|g" $2 >$TMP_FILE
+
+#
+#  MacOS's groff is missing .UR and .UE support, which makes
+#  URL completely disappear from man pages.  We need to strip
+#  those codes out when building for MacOS
+#
+if [ $(uname -s) = "Darwin" ]; then
+    sed -e "s|.UR ||g" $TMP_FILE | sed -e "s|.UE||g" > $3
+else
+    cp $TMP_FILE $3
+fi
+
+rm -f $TMP_FILE


build-aux/mangen.sh
```diff
@@ -11,4 +11,19 @@ if [ $# -lt 3 ]; then
...
-sed -e "s|@VERSION[@]|$1|g" $2 >$3
+TMP_FILE=/tmp/mangen.$$
+
+sed -e "s|@VERSION[@]|$1|g" $2 >$TMP_FILE
+
+#
+#  MacOS's groff is missing .UR and .UE support, which makes
+#  URL completely disappear from man pages.  We need to strip
+#  those codes out when building for MacOS
+#
```
rewolff: First. Multiple sed-commands can be added after multiple -e options. 
sed -e s/one/two/g -e s/three/four/g
Second: How about: 
RMURUE=""
if [ $(uname -s) = "Darwin" ] ; then
  RMURUE='-e "s|.UR ||g" -e "s|.UE||g"'
fi
and then the original: 
 sed -e "s|@VERSION[@]|$1|g" $RMURUE $2 >$3

This removes the tmpfile, and the unnecessary copy

rewolff:
I added a comment in the code.

build-aux/mangen.sh
matt-kimball: Quoted arguments with embedded spaces won't parse the way sed expects them to when you do shell variable expansion that way.  (There will be a '`"s|.UR`' argument followed by a '` ||g"`' argument, and sed won't like that.)

I think you'd have to do something like this to get the shell to parse them the way you'd expect:

```
echo sed -e \"s|@Version[@]|$1g\" $RMURUE $2 | sh >$3
```

That seems harder to understand to me, but I can do it that way if you prefer.

rewolff: yea. That stuff I've debugged now and committed and pushed a rewrite. Can you check it works as expected for you? (for debugging I removed the if(darwin).  )

matt-kimball: Yes, your rewrite works as expected for me.

Issue #308, opened 2019-07-28
Feature Request for Split Mode


eakteam:
It would be great to have an option for -p (Split mode)  to display Double values, it is really needed in some cases and makes the back-end very difficult to parse raw mode to calculate everything by self. Also it would be great to add current ping time into it.

For e.x current split mode : 

```
<pos> <host> <loss%> <rcvd pckts>(Int) <sent pckts>(Int) <best>(Int) <avg>(Int) <worst>(Int) 
```

Very useful split mode if it can be done in following : 

```
<pos> <host> <loss%> <rcvd pckts> <sent pckts> <curr>(Double) <best>(Double) <avg>(Double) <worst>(Double)
```

eakteam:
It is currently displaying like above example and i don't know why it is not showing numbers in Double format, it is only shows Int values.

```
2 10.0.51.26 0 7 7 18 29 41
3 10.128.92.13 0 7 7 44 53 60
4 10.128.92.14 0 7 7 42 49 57
5 10.128.92.236 0 7 7 43 53 57
6 10.128.92.1 0 7 7 43 49 62
7 10.176.16.134 0 7 7 44 54 68
8 83.224.40.185 0 7 7 43 50 58
9 83.224.46.233 0 7 7 45 53 61
```

eakteam:
The current display modes which i am experiencing after cross-compiling :

**-l (Raw Mode) - ./mtr -l -c 1 8.8.8.8**

```
x 0 33000
x 1 33001
h 1 10.0.51.26
d 1 10.0.51.26
p 1 27540 33001
x 2 33002
h 2 10.128.92.13
d 2 10.128.92.13
p 2 55984 33002
x 3 33003
h 3 10.128.92.14
d 3 10.128.92.14
p 3 63097 33003
x 4 33004
h 4 10.128.92.236
d 4 10.128.92.236
p 4 53002 33004
x 5 33005
h 5 10.128.92.1
d 5 10.128.92.1
p 5 54266 33005
x 6 33006
h 6 10.176.16.134
d 6 10.176.16.134
p 6 54586 33006
x 7 33007
h 7 83.224.40.185
d 7 83.224.40.185
p 7 54465 33007
x 8 33008
h 8 83.224.46.233
d 8 83.224.46.233
p 8 52633 33008
x 9 33009
x 10 33010
h 10 8.8.8.8
d 10 8.8.8.8
p 10 45269 33010
x 11 33011
h 11 8.8.8.8
[...]

PR #310, opened 2019-07-28
Fix/raw dns

### LINKED ISSUE #309: dnsline in raw output shows only ip address
the raw output, outputs the dnsline before the hostname as been resolved.  Looks like this behaviour was introduced in 02a9e6193e5dec01198f7824b7ed92f320b4e7a9

krisl:
output dnsline only if hostname is resolved
fixes #309

### DIFF
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -277,7 +277,7 @@ char *dns_lookup2(
         if (r->name)
             return r->name;
         else
-            return strlongip(ctl, ip);
+            return NULL;
     } else {
         r = xmalloc(sizeof(struct dns_results));
         memcpy(&r->ip, ip, sizeof(r->ip));
@@ -289,7 +289,7 @@ char *dns_lookup2(
         if (rv < 0)
             error(0, errno, "couldn't write to resolver process");
     }
-    return strlongip(ctl, ip);
+    return NULL;
 }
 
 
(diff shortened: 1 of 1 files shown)


krisl:
@rewolff can I get some feedback on this PR please
Without this change, the raw mode never outputs a dnsline with a hostname, always just an ip address.

you can see an example in the raw output here https://github.com/traviscross/mtr/issues/308#issuecomment-515730925
every `h` line is followed by an identical, unresolved `d` line

rewolff:
The original idea was to have the low-level stuff do the translation: so we have  a 4-byte integer and the low level stuff converts that to something printable. Either 8.8.8.8 or dns3.google.com, depending on the state of the resolver and the options specified. That makes the higher level code simpler: Simply call the conversion routine and display the results unquestioned. So that's why I don't like the 
 "check-for-null and do something else if that happens".   

Your change never implements the check-for-null that I expected to see. Why isn't it necessary?

krisl:
I modified the function `dns_lookup2`.  There are two callers of it.  One from `dns_lookup` and the other from `raw_rawping`.
The problem is that at the time that `raw_rawping` is called, the dns resolution has not completed, so will always return just the ip address.
`raw_rawping` already has "check-for-null and do something else if that happens" logic, so I added similar logic to the other caller and allowed the function to return NULL.

The main issue being that the callers dont know when the name is resolved.  It looks like they rely on calling this function until something good comes out of it, essentially its polled.  I think the "right fix" here is to have the code output the resolved name at the time that it happens, and I have another branch that im working on to do that, but its a bigger job and i'd like to get much more discussion about it first.

At this time, its just broken.  It never resolves the name.  And this is the simplest change that I could think of to address the issue.  It only alters the behaviour for the one caller - raw mode.  And the other is preserved by moving the logic to that caller (the ternary function)

rewolff:
Thanks for the explanation.

Issue #313, opened 2019-08-05
Error decoding localhost address on macOS


mcdado:
![Screenshot of the error: "Error decoding localhost address. mtr: Failure to start mtr-packet: Invalid Argument"](https://user-images.githubusercontent.com/898057/62471732-92135600-b79d-11e9-9e7f-4d5d612199aa.png)

> Error decoding localhost address.
> mtr: Failure to start mtr-packet: Invalid Argument

Installed via homebrew, with /usr/local/sbin in $PATH. Running with zsh but reproduced in bash too.

rewolff:
Your "mtr" binary needs to execute "mtr-packet" to do the low-level stuff. The error means it can't find it. Can YOU find mtr-packet? Where is it? Is that in your path?

mcdado:
```
sudo mtr-packet 
Password:
Error decoding localhost address
```

rewolff:
Hi, 
I've produced a patch, that will hopefully explain better what goes wrong. 
I've had to upload it with the "log" extension because ".diff" is not accepted by github. patch -p1 < probe_err.log  should patch your sources. 

[probe_err.log](https://github.com/traviscross/mtr/files/3468499/probe_err.log)

cebor:
I have the same issue since the last mtr update via brew.

I also added some addresses for 127.0.0.1 to my /etc/hosts, maybe this is the issue?

rewolff:
can you check the patch and try to rebuild?

mcdado:
```
➜  mtr git:(master) git apply ~/Downloads/probe_err.diff
```

```
➜  mtr git:(master) ✗ ./bootstrap.sh && ./configure && make
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking how to run the C preprocessor... gcc -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
[...]

PR #315, opened 2019-08-06
probe: fix find_source_addr

lembacon:


packet/probe.c
```diff
@@ -323,7 +323,7 @@ int find_source_addr(
        anything to the port.
      */
     *sockaddr_port_offset(&dest_with_port) = htons(1);
-    len = sockaddr_addr_size(&dest_with_port);
+    len = sockaddr_size(&dest_with_port);
```
lembacon: `sockaddr_addr_size` returns size of `struct in_addr` or `struct in6_addr`, while `sockaddr_size` returns size of `struct sockaddr_in` or `struct sockaddr_in6`, `connect` expects the latter.

packet/probe.c
```diff
@@ -364,7 +364,7 @@ int find_source_addr(
        Zero the port, as we may later use this address to finding, and
        we don't want to use the port from the socket we just created.
      */
-    *sockaddr_port_offset(&srcaddr) = 0;
+    *sockaddr_port_offset(srcaddr) = 0;
```
lembacon: `srcaddr` is already the pointer, however it will lead to segmentation fault.

lembacon:
This should fix #313.

PR #316, opened 2019-08-06
asn: always create ipinfo cache

zorun:
We currently have two ways of displaying ASN and other IP information:

1) via a command-line switch (-z or -y)

2) interactively by pressing "z" or "y"

For some reason, we were only creating a cache in the first case, leading
to a high number of repeated DNS queries in the second case.  Fix this by
always creating the cache.

### DIFF
--- a/ui/asn.c
+++ b/ui/asn.c
@@ -313,11 +313,9 @@ int is_printii(
 void asn_open(
     struct mtr_ctl *ctl)
 {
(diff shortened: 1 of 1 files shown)


meingtsla:
Also see pull request #166.

zorun:
Oh! This is an old bug then...  Your patch is even better because you correctly destroy the hashtable in all cases.

@rewolff please merge #166 instead then!

Issue #314, opened 2019-08-06
Feature Request - migrate from gtk2 to gtk3/gtk4 or qt


peterhoeg:
The GUI is pretty simple but with the move from a deprecated gtk2 to either gtk3/4 or Qt, we would get hidpi and wayland support (amongst others) which would be great.

krisl:
resolved by #318

rofl0r:
can you keep backwards compatibility to GTK+2 please ?

rewolff:
@rofl0r You're in luck. I was very much in favor of your request and pushed @krisl to make it that way. :-) Glad to see that others see a need. Can you explain why it is important to you? In this case, we're already beyond the point. The code has been upgraded to use gtk3 when available, but continues to run on gtk2 when THAT is available.

rofl0r:
in sabotage linux we only use GTK+2 apps in the default desktop, because it's the only widely used toolkit written in C that has acceptable bloat levels & is stable and with a consistent look and feel. also having desktop apps that use distinct toolkits, additional memory is needed to load all the libs of those toolkits, which makes memory usage explode if you have stuff that use utterly bloated C++ stuff like QT4/5.

rewolff:
Thank you very much for your explanation. It would be appreciated if you double check my work that mtr still compiles and works with gtk2. (check current git-master).

matt-kimball:
Do people actually use the GTK+ interface?

A while ago I was going to suggest getting rid of it and going only with the curses interface, since the GTK+ interface is badly designed and lacks much of the functionality present in the curses UI.  But I guess if people are really using it, we can keep it.

I say this as the person that wrote the original version of the GTK+ UI a long time ago.

peterhoeg:
> Do people actually use the GTK+ interface?

Absolutely - and that is said as someone who pretty much lives in the terminal.

rofl0r:
@rewolff i built master and got the following odd output
```
checking for GTK... no
checking for GTK... yes
```
but apart from that the build worked fine.

rewolff:
That's an "autoconf" issue. I'm bad with autoconf. ```
AS_IF([test "x$with_gtk" = "xyes"],
  [PKG_CHECK_MODULES([GTK], [gtk+-3.0],
    [AC_DEFINE([HAVE_GTK], [1], [Define if gtk+ library available])]
      [AC_DEFINE([HAVE_GTK3], [1], [Define if gtk+-3.0 library available])],
[...]

PR #317, opened 2019-08-08
Gtk3

krisl:
addresses #314

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,11 +86,11 @@ AC_CHECK_LIB([m], [floor], [], [AC_MSG_ERROR([No math library found])])
 
 # Find GTK
 AC_ARG_WITH([gtk],
-  [AS_HELP_STRING([--without-gtk], [Build without the GTK+2.0 interface])],
+  [AS_HELP_STRING([--without-gtk], [Build without the GTK+3.0 interface])],
   [], [with_gtk=yes])
 AS_IF([test "x$with_gtk" = "xyes"],
-  [PKG_CHECK_MODULES([GTK], [gtk+-2.0],
-    [AC_DEFINE([HAVE_GTK], [1], [Define if gtk+-2.0 library available])],
+  [PKG_CHECK_MODULES([GTK], [gtk+-3.0],
+    [AC_DEFINE([HAVE_GTK], [1], [Define if gtk+-3.0 library available])],
     [with_gtk=no])
 ])
 AM_CONDITIONAL([WITH_GTK], [test "x$with_gtk" = xyes])
--- a/ui/gtk.c
+++ b/ui/gtk.c
@@ -274,22 +274,22 @@ static void Toolbar_fill(
 
     Button = gtk_button_new_from_stock(GTK_STOCK_QUIT);
     gtk_box_pack_end(GTK_BOX(Toolbar), Button, FALSE, FALSE, 0);
-    g_signal_connect(GTK_OBJECT(Button), "clicked",
+    g_signal_connect(G_OBJECT(Button), "clicked",
                      G_CALLBACK(Window_destroy), NULL);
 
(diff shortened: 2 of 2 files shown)


rewolff:
Hi, It looks nice and simple: Good! 
I have one worry. And that's that some corner where MTR used to compile now suddenly won't because the platform has GTK2 and not 3. This could be something like "macos" or "linux-environment on windows", where a big vendor demonstrates: "look you can compile a linux application" by compiling the one application that they worked on and then call it quits. 

The "G_OBJECT" vs "GTK_OBJECT" would easily be fixed: 
```
#if COMPILING_FOR_GTK2
#define G__OBJECT GTK_OBJECT
#else 
#define G__OBJECT G_OBJECT
#endif 
```
and this somewhere in a header. The actual code looks almost the same (no ugly #ifdefs inside the actual code). But some of the others would be more difficult. 

What do you think? Is this an issue? Can we drop GTK2 support just like that?

krisl:
@rewolff  Cool, thanks

Those first two commits are valid even for GTK2, I can make a separate PR if you like?

And the third commit will then switch to using GTK3, but with warnings.  The next three commits just fix the warnings.
So one possibility would be just to allow GTK3 to be compiled with warnings.  Maybe you want a different `configure` option for that, eg `--with-gtk2` vs `--with-gtk3`?  or just to use the highest version found.

Does anyone know of other projects that support multiple GTK versions?

More than that, idk.  I suppose we can expect later versions only to diverge further.  What is the policy for supporting which gui frameworks?  We could make separate GTK2 and GTK3 versions of functions for where they differ significantly.  Or ultimately different files altogether.

I suppose the easiest thing to do, is for those platforms where GTK3 is not available (do such environments exist?), then they no longer will be complied with GTK support.

rewolff:
No separate pull request required for now. First think about it for a second. And I'm thinking aloud in the hopes of getting feedback for when I think/say stupid things. 

If the differences are in compilation warnings or stuff like that, then I'd prefer GTK3 to be warning-less and leave those warnings for those who have to work with gtk2. 

Can I force a compilation with "only gtk2"? I have a machine that is quietly chugging along on ubuntu 14.04 but alas it doesn't have gtk at all. And just apt-get installing it won't work anymore.

krisl:
how about this approach instead? https://github.com/traviscross/mtr/pull/318

@rewolff I'm not quite following what you said about "only gtk2".  If gtk isn't found at all, then it will just build without it. no?

rewolff:
I have a relatively recent raspberry running somewhere. I explicitly installed libgtk-2.0-dev. I cloned the public repo, compiled and it works with Gtk support. Then after pulling your patch, I get "gtk     :no" detected in configure. 

I then reversed the "gtk-3.0 required" in configure.ac and I get two compile errors. Changing the "box new" to hbox_new and vbox_new as seemed to be required on the "orientation" argument and removing the call to gtk_menu_popup_at_pointer, I get a working, but ugly mtr. 
The ugly might be my guessing wrong for hbox/vbox. 

I'd say that keeping gtk2 compatiblity is doable. The configuration asks for "at least 2.0" will probably default to 3.0 when available. And a simple header that defines  gtk_hbox_new and gtk_vbox_new as gtk_box_new for gtk3 ONLY and for all I care it defines gtk_menu_popup_at_pointer as "nothing", then we have this working on both versions. 
If it is tons of work I'd say leave it, but it seems easy enough. 
Note that my code-style preference is to have the code READ as if it is written for the latest stuff, but with the hbox/vbox stuff we have to compromise: the compatiblity defines are easyest when we just keep the old code but insert a define to rewrite it on the fly for the new GTK.

krisl:
cool, thanks for the explanation.
closing as the gtk3 with fallback to gtk2 branch is merged #318

PR #318, opened 2019-08-08
Gtk3 with fallback

krisl:


krisl:
as an alternative to #317, will try to build with GTK2 if GTK3 is not found, uses two `#ifdef`s to ensure no warnings when compiling for GTK3

rewolff:
Ah!!! Yes, Almost what I wanted.... Let me try and pull it, fix it and then merge... :-)

rewolff:
merged.

PR #322, opened 2019-10-10
sockaddr related cleanups

commonism:
remove redundant functions and try to simplify/improve readbility

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -55,6 +55,7 @@ mtr_SOURCES = ui/mtr.c ui/mtr.h \
               ui/select.c ui/select.h \
               ui/utils.c ui/utils.h \
               packet/cmdparse.c packet/cmdparse.h \
+              packet/sockaddr.c packet/sockaddr.h \
               ui/mtr-curses.h \
               img/mtr_icon.xpm \
               ui/mtr-gtk.h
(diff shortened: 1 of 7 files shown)


rewolff:
Hi, 
You've removed  a bunch of casts to " (void *) " . I'm guessing those were there due to otherwise we're getting warnings about type conflicts on those. 

I like code to compile at a high warning-level without any warnings. (like those). 

IF it is possible that we'd get a warning on SOME system that mtr supports... Are you sure we've done the types correctly? i.e. that we're correctly following some standard about how to do such things? 

(Getting warnings is acceptable, as long as we can prove we're doing the right thing.).

commonism:
I changed the signature of addrcmp to accept "void *" instead of "char *", so there is no need to cast the arguments when using it.
Improving readability. This does not interfere with the compiler/warnings, I like the compiler to warn as well.

https://github.com/commonism/mtr/blob/283dc75726ef115d6ce0b632a9e32d05df4f1efb/ui/net.h#L119-L123

rewolff:
Hi, 
I'm scanning this, and I'm thinking: Looks good, merge! But the merge button is not there. It seems I thought precisely the same about a month ago and merged it already. If there is anything missing I think you'll have to create a new pull request.

PR #336, opened 2020-02-16
show mpls information in raw output

ym:


ui/net.c
```diff
@@ -259,7 +259,7 @@ static void net_process_ping(
             memcpy(&nh->addrs[i], addrcopy, sockaddr_addr_size(sourcesockaddr));
 
             nh->mplss[i] = *mpls;
-            display_rawhost(ctl, index, (void *) &(nh->addrs[i]));
+            display_rawhost(ctl, index, (void *)&(nh->addrs[i]), (void *)&(nh->addrs[i]));
```
vincentbernat, line 262: This doesn't seem right. The last argument should be `&nh->mpls`. Is the cast needed? I am removing these casts in #389.

PR #337, opened 2020-02-21
Destination hostname displayed in curses.

priddle:
I run multiple mtr's in different windows to monitor several networks. I wanted to be able to quickly identify which window belonged to which network.


priddle:
fixed with issue #345

PR #355, opened 2020-07-09
report: json: reworked with libjansson

ltWolfik:
I reworked JSON report to use libjansson. Consider such solution more robust.
I noticed that fixes JSON output when no hosts:
<details>
  <summary>Screenshot with comparison</summary>
  drawing
</details>
Also, it changes the following:

- report.mtr.{tests, tos} now output as number;
- report.hubs[].count now output as number;
- the precision of real numbers differs;

<details>
  <summary>Screenshot with comparison</summary>
  drawing
</details>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -92,8 +92,8 @@ ASAN_CFLAGS += -fsanitize=address
 endif
 
 mtr_INCLUDES = $(GLIB_CFLAGS) -I$(top_builddir) -I$(top_srcdir)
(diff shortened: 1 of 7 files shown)


rewolff:
What happens if libjansson isn't installed?

ltWolfik:
I supposed libjansson should be mandatory dependence. But if it's critical - do it conditional dependence (like gtk and ncurses)?

rewolff:
Yeah. I would prefer to keep things in line with other output modules: If the library for that feature is not available, that feature won't be built.

ltWolfik:
I did Jansson to be optional dependence. Can you accept the PR?

PR #368, opened 2020-09-22
ncurses: Link against libtinfo as needed

jer-gentoo:


acoul:
Greetings,

I am a gentoo, arch & void(musl) linux user. 

although the above patch does solve the compile issue on my Gentoo system, it creates o non-ncurses mrtg on my other two systems

hhoffstaette:
Just an explicit note that since commit https://github.com/traviscross/mtr/commit/25fc5274ef49465cf0c3236ebf2445ddf49f4647 (aka 0.96) this PR will create problems, since it's necessary to link against tinfo**w**.

Issue #250, opened 2018-03-31
Option -a/--address does not work.


bd46:
Option "-a|--address" does not work in package "mtr-tiny" (build without X11) on Debian Linux, version  0.92-1. Running under strace shows that program does NOT call bind(2).

Breif examination of source code from Github reveals that -a parameter value is assigned to ctl->InterfaceAddress, then is validated in net.c::net_validate_interface_address(), and never used later. Probably some piece of code was lost.

In previous mtr versions option "-a" worked as expected.

bd46:
After bug submission I found the same bug as #232, sorry for duplicaion. Testcase is quite simple: multihomed host (several interfaces with different addresses), mtr sent packets via default route regardless of -a option.

rkandilarov:
Just to confirm that I also came in practice to the same bug `-a source_IP` not working with ICMP, it sends the packets with the proper source_IP but trough the default interface (gateway) not the one associated with the provided source IP. With `--tcp` it works as expected.

rewolff:
Test case. Normal PC, `ifconfig eth0:1 <a second IP addres>` 
Now `ping -I <second ip adress>` a host and see that the second IP address is being used. 
Now `mtr -a <second ip address>` that host and see that the second IP address is being used. 

The -I on "ping" hints at using an INTERFACE as the source, but I don't think that can be done with the networking-interface on Unix-like machines. the -a option on mtr explains better what it does: The source **address** is set. The kernel then often decides to use the interface that you meant to use, but if it doesn't... well mtr can't do much about that.

bd46:
On Thu, Aug 02, 2018 at 11:36:19PM -0700, Roger Wolff wrote:
>    Test case. Normal PC, ifconfig eth0:1 <a second IP addres>
>    Now ping -I <second ip adress> a host and see that the second IP address
>    is being used.
>    Now mtr -a <second ip address> that host and see that the second IP
>    address is being used.
> 
>    The -I on "ping" hints at using an INTERFACE as the source, but I don't
>    think that can be done with the networking-interface on Unix-like
>    machines. the -a option on mtr explains better what it does: The source
>    address is set. The kernel then often decides to use the interface that
>    you meant to use, but if it doesn't... well mtr can't do much about that.

 To set ip-address of source packets application MUST(!) do bind(2) syscall.
 For check run "strace -e bind -o mtr.trace mtr -a <src_ip> <dst_ip>" and
 look into resulting mtr.trace, you will see that mtr does not call bind().

[...]

PR #371, opened 2020-09-25
Fixes to make it behave more like stable 0.87

totoCZ:
* Fix for #232  #250 
* Reverted d2552ca because it breaks MPLS and hosts start jumping up and down in the interactive display

I'm not certain about some things in deconstruct_unix (why (struct ICMPHeader *) (ip + 1) in handle_received_ip4_packet? It should be the same for both sockets.

But I don't have any more time/energy to work on improving it further.

### DIFF
--- a/packet/construct_unix.c
+++ b/packet/construct_unix.c
@@ -71,19 +71,6 @@ uint16_t compute_checksum(
     return (~sum & 0xffff);
 }
 
-/*  Encode the IP header length field in the order required by the OS.  */
-static
-uint16_t length_byte_swap(
-    const struct net_state_t *net_state,
-    uint16_t length)
-{
-    if (net_state->platform.ip_length_host_order) {
-        return length;
-    } else {
-        return htons(length);
-    }
-}
-
 /*  Construct a combined sockaddr from a source address and source port  */
 static
 void construct_addr_port(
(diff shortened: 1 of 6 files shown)


rewolff:
> Revert "Change TTL dynamically to adjust for path changes" ￼…
You want to revert some "improvements" that others are working on. You need to provide a better argument than: "I don't like it". For example a case where it did something wrong, so that we can go forward towards "the best of both worlds".

TomHetmer:
Wrong, I don't need to provide anything.

Anyhow, you better read why it's broken.

>because it breaks MPLS and hosts start jumping up and down in the interactive display

rewolff:
OK. I don't know how to reproduce the problems you're seeing.

TomHetmer:
Easy, run mtr with -e, ie.

 ./mtr upc.cz -e

Passes through aorta.net with MPLS, no MPLS labels are displayed.

Run mtr with -u

./mtr upc.cz -u

Unlike versions before that commit multiple hosts on each line start jumping up and down instead of staying at the fixed place making it super hard to read.

acoul:
FWIW, I have tested the above patches against mtr-0.94 (with & without the Revert "Change TTL dynamically to adjust for path changes") but issue https://github.com/traviscross/mtr/issues/232 is still present

totoCZ:
patch has been merged
https://github.com/traviscross/mtr/pull/484

PR #384, opened 2021-02-01
truncate to lower 16 bits

### LINKED ISSUE #383: ui/mtr.c:696:21: runtime error: left shift of 2277731 by 16 places cannot be represented in type 'int'
When MTR 0.94 is built with libasan (e.g. on RHEL/CentOS 8), the startup throws:

```
ui/mtr.c:696:21: runtime error: left shift of 2277731 by 16 places cannot be represented in type 'int'
```

rpaaron:
fixes #383

### DIFF
--- a/ui/mtr.c
+++ b/ui/mtr.c
@@ -693,7 +693,7 @@ static void init_rand(
     struct timeval tv;
 
     gettimeofday(&tv, NULL);
-    srand((getpid() << 16) ^ getuid() ^ tv.tv_sec ^ tv.tv_usec);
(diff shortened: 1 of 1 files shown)


rewolff:
I think this is the wrong way to fix this. 

The idea was to get "more-or-less-random" bits in all the bits of the value we pass to srand. 
When getuid and pid both were 16 bits, shifting one of them by 16 bits seemed like a way to put "something" in the higher order bits. Now that pid has gone 32-bits, I propose to just use all the bits of pid and shift the uid by 16.

Issue #378, opened 2020-11-07
GTK3 UI doesn't refresh target when launched without a target


mikelolasagasti:
If mtr is launched with `mtr -g www.google.com` and then change the target in the UI, it works and data of the new target is reported.

If mtr is launched with `mtr -g` and then change the target in the UI interface, it doesn't change target and reports nothing.

Tested with 0.94 and master in Fedora 33.

Example: https://www.youtube.com/watch?v=h-tcjw3f3Jo

robert-scheck:
This issue is not GTK3 specific, but reproducible with GTK2, too

krisl:
`./mtr localhost` has this problem, also
`./mtr 127.x.x.x` also has this problem
`./mtr <some_other_local_interface` appears to be ok.
Not sure if it just a gtk issue, or perhaps its not even sending the probes

robert-scheck:
@krisl, thank you for the pointer!

I've now analyzed this, but I'm not sure how to fix it: The problem is that `net_reopen()` does neither reinitialize the source address itself nor the source address family (which `net_open()` initially does both). This means, if `mtr -g 127.0.0.1` is run, `mtr` tries to re-use the source address 127.0.0.1 for the next entered (non-loopback) destination, which obviously does not work (because you can not reach 1.1.1.1 from 127.0.0.1). However, as nowadays on most Linux distributions `mtr -g localhost` is equivalent to `mtr -g ::1`, `mtr` tries to re-use the source address ::1 for the next entered (non-loopback and/or possibly even IPv4) destination, which obviously does not work as well (because you can reach neither 2606:4700:4700::1111 nor 127.0.0.1 from ::1).

So, from my point of view, a complete testing scenario (once a fix is proposed) would be:

1. `mtr -g 127.0.0.1`, then switch in the GUI to `1.1.1.1` (or any other non-loopback IPv4 address)
2. `mtr -g 127.0.0.1`, then switch in the GUI to `2606:4700:4700::1111` (or any other non-loopback IPv6 address)
3. `mtr -g ::1`, then switch in the GUI to `2606:4700:4700::1111` (or any other non-loopback IPv6 address)
4. `mtr -g ::1`, then switch in the GUI to `1.1.1.1` (or any other non-loopback IPv4 address)

krisl:
@robert-scheck awesome sleuthing!  I've made a pull request at https://github.com/traviscross/mtr/pull/393

robert-scheck:
[...]

PR #393, opened 2021-04-03
Fix changing source

rpaaron:
Fixes https://github.com/traviscross/mtr/issues/378

I moved the "starting a new traceroute" part of `net_reopen` to it single caller in `gtk.c`
Then moved the common parts of `net_open` to `net_reopen` including the initialisation of the source addresses, and now call `net_reopen` from `net_open`

### DIFF
--- a/ui/gtk.c
+++ b/ui/gtk.c
@@ -249,6 +249,7 @@ static gint Host_activate(
     addr = dns_forward(gtk_entry_get_text(GTK_ENTRY(entry)));
     if (addr) {
         net_reopen(ctl, addr);
+        net_send_batch(ctl);
         /* If we are "Paused" at this point it is usually because someone
            entered a non-existing host. Therefore do the go-ahead... */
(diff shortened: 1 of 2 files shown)


robert-scheck (commented):
I am sorry, but this patch is incomplete: Running `mtr -g ::1` and then switching to `8.8.8.8` in the GUI does *not* work, but aborts with `mtr: Address not available`

rewolff:
Yeah.... work needs to be done to be able to switch address families, right?

robert-scheck:
Exactly!

rewolff:
The annoying thing is that starting without a target and then specifying the target in the gui is a way to work with mtr, but if it chooses the AF wrong with the default target that would be annoying.

robert-scheck:
No default target/destination leads to `AF_INET6` on most systems, because localhost resolves to ::1. But as the GUI allows to switch the target/destination (while the TUI doesn't), the address family needs to be re-evaluated with each target/destination change.

PR #394, opened 2021-04-03
tell dns process which family we want resolved

krisl:
[edit] was addressing 378
In particular addresses the case where changing from one address family to another. ie when launching mtr with an ipv6 address, then in the gtk interface changing to an ipv4 address.

@robert-scheck I currently cannot test this, are you able to see if it addresses your use case?

### DIFF
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -124,7 +124,7 @@ static void set_sockaddr_ip(
 }
 
 void dns_open(
-    sa_family_t family)
+    void)
 {
     int pid;
 
@@ -173,7 +173,8 @@ void dns_open(
 
                 buf[strlen(buf) - 1] = 0;       /* chomp newline. */
 
-                longipstr(buf, &host, family);
+                sa_family_t family = (buf[0] == '4') ? AF_INET : AF_INET6;
+                longipstr(buf +1, &host, family);
                 set_sockaddr_ip(family, &sa, &host);
                 salen = (family == AF_INET) ? sizeof(struct sockaddr_in) :
                     sizeof(struct sockaddr_in6);
@@ -256,7 +257,7 @@ char *dns_lookup2(
     ip_t * ip)
 {
     struct dns_results *r;
-    char buf[INET6_ADDRSTRLEN + 1];
+    char buf[INET6_ADDRSTRLEN + 2]; // af_byte + addr + null
     int rv;
 
     r = findip(ctl, ip);
@@ -270,7 +271,8 @@ char *dns_lookup2(
         r->name = NULL;
         r->next = results;
         results = r;
-        snprintf(buf, sizeof(buf), "%s\n", strlongip(ctl->af, ip));
+        char ip4or6 = (ctl->af == AF_INET) ? '4' : '6';
+        snprintf(buf, sizeof(buf), "%c%s\n", ip4or6, strlongip(ctl->af, ip));
         rv = write(todns[1], buf, strlen(buf));
         if (rv < 0)
             error(0, errno, "couldn't write to resolver process");
--- a/ui/dns.h
+++ b/ui/dns.h
@@ -23,7 +23,7 @@
 /*  Prototypes for dns.c  */
 
 extern void dns_open(
-    sa_family_t family);
+    void);
 extern int dns_waitfd(
     void);
 extern void dns_ack(
--- a/ui/mtr.c
+++ b/ui/mtr.c
@@ -858,7 +858,7 @@ int main(
         }
 
         lock(stdout);
-        dns_open(ctl.af);
+        dns_open();
         display_open(&ctl);
 
         display_loop(&ctl);


robert-scheck:
@krisl, while it seems to solve launching with an IPv6 address and switching then to an IPv4 address, it for sure breaks all reverse DNS lookups (in the trace), where the following message is thrown once or multiple times:

```
mtr: dns_ack: Couldn't find host 8000::f9a8:6760:0:0
```

Launching directly with IPv4 address, no messages appear and reverse DNS lookups are working.

krisl:
hmm yes, because `dns_open` forks a new process early on with its original copy of `ctl` that it keeps using to check the address family.
What can we do?  
- restart the dns process every net_reopen?
- in the dns process, dont use `ctl` for address family, but parse every request individually to determine family
- make a way of communicating to the dns process that the address family has changed?
- run separate v4 and v6 dns lookup processes?

I'm thinking treating every request for dns looking as independent, and unrelated to the "session", will be the least invasive and most reliable going forward

robert-scheck:
Ush…while it looked in the beginning like a semi-simple "mixup of loopback vs. IPv4 vs. IPv6"-thing, this goes now quite far. I'm happy to test further PR etc., but I don't treat myself capable of actually implementing that.

krisl:
ok, least invasive thing I could think of was to add explicitly to the request to the dns process, whether we want ipv4 or ipv6 address resolution - by prefixing the message in the pipe with a character "4" or "6".

specifically this last commit [tell dns process if we want 4 or 6](https://github.com/traviscross/mtr/pull/394/commits/d7388c7b7dd3e773f13a88c3b0fc62eb83f6dbf6)

the approach feels ok to me, the implementation.. subject to review
thoughts?

robert-scheck:
Wow…technically, this implementation seems to work properly!

There is one minor thing, which is strange: `mtr -g google.de` uses IPv6 (yes, I'm dual-stacked here), while `mtr -g ::1` and then switching to google.de uses IPv4 (rather IPv6). My expectation would be, that like for the TUI variant, GUI should prefer IPv6 over IPv4 (if IPv6 is available), like RFC 6724 says.

krisl:
> There is one minor thing, which is strange: `mtr -g google.de` uses IPv6 (yes, I'm dual-stacked here), while `mtr -g ::1` and then switching to google.de uses IPv4 (rather IPv6). My expectation would be, that like for the TUI variant, GUI should prefer IPv6 over IPv4 (if IPv6 is available), like RFC 6724 says. Note that I did _not_ restrict the address family using `-4` or `-6` in my tests.

I think what is going on there is that the method the GUI uses to parse the host, and the method the TUI uses are different.  GUI uses the method `dns_forward` which is just a wrapper around the deprecated `gethostbyname`.    And the TUI uses `get_hostent_from_name` uses the newer `getaddrinfo`.

I think in total we have discovered four issues here.

1.  Cannot change from loopback interface to non-loopback.  addressed in #393
2. Cannot change from one address family to another.  Addressed in this PR
3. DNS resolution fails when switching from one address family to another.  Also addressed here, but perhaps should be separated out
4. GUI behaves differently from TUI.  Still to be fully investigated

rpaaron:
fixes https://github.com/traviscross/mtr/issues/396

rewolff:
@krisl when it's ready to merge, can you remove the "draft" status?

robert-scheck:
Is it intended that since the force-push, the IPv6 reverse DNS lookups are not working anymore?

rpaaron:
> Is it intended that since the force-push, the IPv6 reverse DNS lookups are not working anymore?

no, that would be a mistake somewhere if that is happening.  it should work for both ipv6 and ipv4, but I don't have any immediate ability to check ipv6.

robert-scheck:
> no, that would be a mistake somewhere if that is happening. it should work for both ipv6 and ipv4, but I don't have any immediate ability to check ipv6.

Most simple reproducer that I found is `mtr -g ::1` (which shows ::1 rather localhost), where `mtr -g 127.0.0.1` shows localhost.

rpaaron:
> Most simple reproducer that I found is `mtr -g ::1` (which shows ::1 rather localhost), where `mtr -g 127.0.0.1` shows localhost.

thanks!  looks like its just that last commit that's doing it.   revert that and its back to resolving both

robert-scheck:
Yes, confirming that commit 941c98a breaks it.

rpaaron:
> Yes, confirming that commit [941c98a](https://github.com/traviscross/mtr/commit/941c98a05f46c626745fdfaad5149a618e91102c) breaks it.

thanks for confirming.  Yeah that commit is pretty hacky. Its experimenting with just writing raw bytes into the pipe between the processes.  But the pipe reading is still done "textwise" with `fgets`.  `AF_INET6` is 0x0A, which is ascii control `STX`, start-of-text.  Causing two short, and now corrupted, reads of the pipe.

rewolff:
Please don't use raw binary when it isn't strictly necessary. On this project we're going for maintainability (i.e. readable concise code) and ease-of-debugging. The nonprintable stuff should be the newlines between the lines as a separator and maybe a tab between fields (use a space whenever possible, but when the fields can have spaces and commas but not tabs, a tab might be acceptable). 

I've started writing a "logger" that will log the traffic being sent in a pipe between two programs. Once that is done, reading the log is much more comfortable with only ascii moving between the two.

krisl:
fixes #396

fun thing I discovered, `AF_INET6`, and others presumably, are not the same values on all systems.  On my linux system is 0x10, on mac 0x30.

ok, simplest non-invasive, readable and changeable-in-future thing I could come up with is just to prefix the address-string-to-be-resovled with... a '4' or a '6'.

robert-scheck:
Okay, commit 8bf96efd57ec9a43fb88cbfe58762182204762b3 works for me regarding IPv4 and IPv6 reverse DNS as expected.

PR #395, opened 2021-04-05
Reset address family

krisl:
fixes https://github.com/traviscross/mtr/issues/378

This should be a set of fairly non controversial changes

With this PR you can start off with `mtr <ip_v6_addr>`, then in the gtk interface, change to a ip v4 address.

Two other issues have been identified in the discussion of https://github.com/traviscross/mtr/pull/394, and I'd like to raise and discuss those separately, and this PR would be a prerequisite for addressing them.


rewolff:
Good cleanup. A change I approve. :-)

PR #411, opened 2021-11-01
ui/curses: always use "%s"-style format for printf()-style functions

trofi:
`ncuses-6.3` added printf-style function attributes and now makes
it easier to catch cases when user input is used in palce of format
string when built with CFLAGS=-Werror=format-security:

    ui/curses.c:765:42:
      error: format not a string literal and no format arguments [-Werror=format-security]
      765 |         mvprintw(rowstat - 1, startstat, msg);
          |                                          ^~~

Let's wrap all the missing places with "%s" format.

### DIFF
--- a/ui/curses.c
+++ b/ui/curses.c
@@ -432,7 +432,7 @@ static void mtr_curses_hosts(
                 attron(A_BOLD);
 #ifdef HAVE_IPINFO
             if (is_printii(ctl))
-                printw(fmt_ipinfo(ctl, addr));
+                printw("%s", fmt_ipinfo(ctl, addr));
 #endif
             if (name != NULL) {
                 if (ctl->show_ips)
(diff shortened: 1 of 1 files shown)


rewolff:
Can you also check recent "issues" IIRC someone also noted a very similar problem recently.

trofi:
Skimmed through a few recent issues and attempted to find similar warning text or format string crashes. Found nothing similar.

rewolff:
Thanks a lot. :-) 
In hindsight it might have been another project where that came by.

samueloph:
I believe it was from Debian:
https://bugs.debian.org/997194

I will backport this fix in case there isn't a new release soon.

Thanks

rewolff:
At some point in time I had a release script that would automate
everything around a release. That way I wouldn't forget to put out the
diff or the tarball or whatever. That broke and then it's taken some
time to fix again. Apart from the script thinking it needed to put out
0.94, I think it works again. I've synced it up and I think it works
again. This makes it way more easy to make a release. (the only issue
now is that the machine the script thinks is the only one that
works... is being decomissioned....)

pabs3:
Thanks for fixing the release script, looking forward to the release.

I hope there is a replacement for the to be decomissioned machine!

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

samueloph:
Message ID: ***@***.***>THan

> Thank you everyone!

I've uploaded 0.95-1 to Debian unstable.

Cheers,


-- 
Samuel Henrique <samueloph>

Issue #351, opened 2020-06-08
IPv6 UDP Probes with fixed ports fail lookup


janrueth:
Hi,

when using fixed src and destination ports and IPv6 and UDP, I do not receive any output.
It seems that the checksum fixup that is performed in IPv4 is not done in v6 (i.e., correcting the payload and actually setting the sequenced checksum).
I see the trace in tcpdump but the checksum is always the same which leads to mtr failing to find the probe when the hop limit exceeded message comes in.

Best
 Jan

23hiro:
@janrueth maybe you would like to confirm this fixes it?

janrueth:
Wow, did not expect to get an answer here.

Just tried it, and it works!

Thanks a lot!

PR #412, opened 2021-11-09
ipv6 udp checksums like ipv4 but with ipv6 pseudoheader

23hiro:
this fixes a problem where both source and destination port are supplied on the commandline,
this lead to packets not showing up any more due to checksum being different.
this was mentioned first on https://github.com/traviscross/mtr/issues/351


23hiro:
@rewolff is there any problem with this, as i haven't heard anything in half a year?

rewolff:
No. Sometimes I'm VERY busy and think: "no time for mtr, tomorrow" and then I'll forget. So reminders are useful when dealing with me. :-) 

I've been too busy the last half year to do a cleanup and then run into this.

23hiro:
thx :)

PR #429, opened 2022-05-13
use addrs for static host ordering in curses

bewing:
Change d2552ca has the side effect of moving hosts around in the
multipath view, as the last host to reply to a packet was set as
addr.  Using addrs[0] istead of addr to draw the first host keeps
the ordering deterministic, in order of first reply.


acoul:
I can confirm that compiling main tree with this patch resolves the  --address issue and both 

`./mtr  sip.nexmo.com  -u`
&
`./mtr  sip.nexmo.com  -e`

work as they should

PR #430, opened 2022-05-13
add --max-display-paths option

bewing:
It is not unknown to have datacenters with over 8 ECMP paths between two
hops.  Expand the maximum number of ECMP hosts stored in addrs to 128,
and add a new option to control the number to display, defaulting to the
old MAX_PATHS of 8

### DIFF
--- a/man/mtr.8.in
+++ b/man/mtr.8.in
@@ -92,6 +92,9 @@ mtr \- a network diagnostic tool
 .BI \-U \ MAX\-UNKNOWN\c
 ]
 [\c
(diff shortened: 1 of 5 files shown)


rewolff:
Hey. Sorry about that, but I'm afraid I saw your other patch first and clicked "merge" on that one. Now this one has conflicts. I think it's easy, but I would like it if you could double check it as you're the one that changed that line.... 
:-)

bewing:
Wouldn't have mattered which merged first, the other would've needed a rebase/manual resolution.  Pushed fixed changeset.

rewolff:
Thanks!

PR #461, opened 2022-12-25
Allow using all UDP ports

### LINKED ISSUE #421: why MinPort?
https://github.com/traviscross/mtr/blob/ec42ba61f77654e8397e6496095634585f90b26d/ui/mtr.h#L69

Why would mtr limit port selection if mtr has the privilege to use the port? There are legit use cases where it's desirable to source from 80/443 (to identify return path issues from a server)

dasJ:
As #421 explains, this is incredibly useful and running mtr as root for this is a cost worth paying. I removed the entire logic to also allow port 0 as Wikipedia states:
> while for UDP, the source port is optional and a value of zero means
no port

Closes #421

### DIFF
--- a/packet/command.c
+++ b/packet/command.c
@@ -225,15 +225,6 @@ bool decode_probe_argument(
         if (*endstr != 0) {
             return false;
         }
-
-        /*
-           Don't allow using a local port which requires
-           privileged binding.
-         */
-        if (param->local_port < 1024) {
-            param->local_port = 0;
-            return false;
-        }
     }
 
     /*  The "type of service" field for the IP header  */
--- a/ui/mtr.c
+++ b/ui/mtr.c
@@ -580,7 +580,7 @@ static void parse_arg(
         case 'L':
             ctl->localport =
                 strtonum_or_err(optarg, "invalid argument", STRTO_INT);
-            if (ctl->localport < MinPort || MaxPort < ctl->localport) {
+            if (MaxPort < ctl->localport) {
                 error(EXIT_FAILURE, 0, "Illegal port number: %d",
                       ctl->localport);
             }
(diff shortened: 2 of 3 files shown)


rewolff:
I don't think this is a good idea. 

Privileged ports are privileged because it requires "root" to open them. Thus a program could say: "hey my buddy computer, I have a user here called John, he wants to run a shell on your computer"... - "are you sure it's john?" - "Yup, I'm sure" - "Ok fine here we go!". That'd be a user-callable setuid-root program on the originating computer that verfiies the local user's identity and then pledges that he's sure to the other computer. The responsibility of checking authorization is moved from one place to another.

To prevent normal users from writing a progam that just says: "Trust me, really, I'm john!", everybody trusts on normal users not being able to open the trusted ports. 

Now MTR needs special network-permissions to send crafted network packets. In the old days that required "root" and that also came with the "open special ports" privilege. Today from your "requires running as root" it seems that your system has decoupled them. 

But still... I can't be sure it is not more similar to or the same on other systems. And I can't be sure you can't make MTR send a packet that's interpreted as "Trust me, I'm john, I just want to execute this small statement there" 

The OS has protections in place, but exceptions to those protections are useful. For example, PING and MTR send "funny packets" on the network That's not allowed by default, but a setuid program like PING will check permissions (don't send anything TOO funny) and then do what it is supposed to do. Similarly mtr should make sure that nothing "too funy" goes out that might trick other computers into doing things they are not supposed to be doing. 

With that said: I don't think it is a good idea to just remove the checks for low ports from MTR.

rewolff:
Privileged ports are "priviledged" for a reason: normal users can't open them and even though we're moving away from such authentication methods, they are far from "no longer used". 

You might check for original uid == 0 , then allow privileged ports. But we elevate privileges to do "funny stuff" and "promise" not to break system integrety. And IMHO that means we don't allow normal users to use a privileged port.

PR #481, opened 2023-07-11
Github actions added to perform lint and compile

darless:
In the test folder, there was lint.sh, but I think flake8 is a better tool. When this PR is reviewed, let's discuss if we can use flake8  in which case I can remove lint.sh.

Added a compile job for linux, this runs compilation as defined in the README, runs a sample mtr to 1.1.1.1 and runs cmdparse.py test. Need documentation on what other tests to run and whether to include the testing in tox instead of running them individually.

Not adding in this PR support for cygwin, macos, or freebsd, as that will need investigation on how to run.

Minor:
- Fixed a typo in configure.ac that used 4 dashes vs 2 for a flag option.

Actions will run when this gets merged in.

The actions ran in my fork from this commit: [MTR github actions run](https://github.com/darless/mtr/actions/runs/5523401612)

Items to discuss in PR prior to merge
- [ ] Within the test directory there is lint.sh which uses pylint. I think flake8 is a better tool and this commit uses that. If this is sufficient, then lint.sh can be removed/modified to use flake8. 
- [ ] Discuss plans for testing, currently a very limited test is done, but we could add cygwin, mac, tests as needed (I would think separate issues for those to add on to the testing. This is only a base testing PR

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,14 @@
+[flake8]
+
+select =
+    E902,
+    E999,
+    S,
+    F,
+
+ignore =
+    F821
+
+per-file-ignores =
+    test/*: S603,S404
+
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,35 @@
+name: Test compilation
+on: [push, pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install flake8
+        run: pip3 install flake8==3.9.2 flake8-bandit==2.1.2 bandit==1.7.2
+      - name: Run flake8
+        run: python3 -m flake8 .
+  compile-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - uses: egor-tensin/setup-gcc@v1.3
+        with:
+          version: latest
+          platform: x64
+      - name: Bootstrap
+        run: ./bootstrap.sh
+      - name: Configure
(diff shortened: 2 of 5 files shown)


darless:
@rewolff let me know what you think about this PR and if you would like to see any changes.

rewolff:
Sorry, kind of busy with work lately. 

Anyway... For the other Unix-like OSes, can't we just enable them "unmodified" from Linux and see what happens? 

Apparently I can install flake8. So now I have that tool. If it replaces "lint", should it report about style discrepancies/errors in the code? and/or report suspicious constructs? e.g. a = b==1?2:5; (that's probably meant to assign 2 or 5 depending on if b equals one, but I think the compiler assigns true (1) or false (0) depending on b equalling two  (1?2:5 evaluates to two)).

darless:
@rewolff The linting is only running on python files which only exist in the test directory. 
```
$ find . -name '*.py'
./test/cmdparse.py
./test/param.py
./test/probe.py
./test/mtrpacket.py
```
The PR contains a flake8 selects:

- E902: Catches potentials for IOError: https://www.flake8rules.com/rules/E902.html
- E999: Caches syntax errors: https://www.flake8rules.com/rules/E999.html
- S: Catches all security issues (provided by flake8-bandit
  - This matches the BXX codes provided by bandit: https://bandit.readthedocs.io/en/latest/plugins/index.html
- F: General errors and validation code: https://flake8.pycqa.org/en/latest/user/error-codes.html

Ignores F821 globally: Undefined name
- Seen in test/probe.py and test/mtrpacket.py (comment that has unicode)
Ignores S603 and S404 in the test directory (deals with usage of subprocess)

@rewolff  Not sure what you mean for the other OSes (unmodified). The reason why i'm saying it would be a different PR, is a different OS (windows-latest or macos) likely has different way of installing the packages that are needed and likely will take much longer to run as an action. I don't think it would be worth the time to add more to this PR than what base testing needs.

PR #484, opened 2023-09-11
Change UDP and ICMP sockets binding to accept a source IP from the -a…

alarig:
… CLI option

Issue: #232


robert-scheck (commented):
This patch enables me to use `--address` and `--interface` successfully on Linux with source/policy based routing, such as two internet upstreams ("two default gateways") and such avoids the previously needed workaround to switch from ICMP to TCP using `--tcp`.

Thank you very much!

Issue #485, opened 2023-09-14
BUG: Interface-specific binding and packet marking broken


flu0r1ne:
I am reporting two distinct issues related to the 'mtr' network diagnostic tool. The first issue (Bug 1) arises when specifying a network interface using the '-I' flag; contrary to expectations, 'mtr' does not bind directly to the specified device but seems to consult the system's routing table instead. This results in packets being sent via an unintended interface, leading to diagnostic inaccuracies. The second issue (Bug 2) occurs when using the '-M' flag to set a firewall mark on packets; despite setting the mark, it appears that 'mtr' does not actually apply it, causing the packets to not follow the expected policy-based routing rules. These two issues hinder the tool's effectiveness in specialized network diagnostics.

To demonstrate these issues, I have created a small proof of concept using Linux network namespaces to demonstrate the bug. Conceptually, it simulates two boxes with two connected interfaces:

```
 ┌──────────────────────┐ 172.30.1.0              ┌──────────────────────┐
 │ MTR-TEST0    VETH01A ├─────────────────────────┤ VETH10A    MTR-TEST1 │
 │                      │               172.30.1.1│                      │
 │                      │                         │                      │
 │                      │ 172.30.2.0              │                      │
 │              VETH01B ├─────────────────────────┤ VETH10B              │
 │                      │               172.30.2.1│                      │
 │                      │                         │                      │
 │                      │                         │                      │
 │                      │                         │                      │
 │                      │                         │                      │
 └──────────────────────┘                         └──────────────────────┘
```

The following tests were conducted on Linux 6.1.51-1-lts using mtr version 0.95.

```bash
#############
### SETUP ###
#############

# Create two network namespaces to simulate a connection between mtr-test0 and mtr-test1
ip netns add mtr-test0
ip netns add mtr-test1

# Create the first pair of linked virtual ethernet interfaces (a)
ip link add veth-01a type veth peer name veth-10a
ip link set veth-01a netns mtr-test0
ip link set veth-10a netns mtr-test1

# Create the second pair of linked virtual ethernet interfaces (b)
ip link add veth-01b type veth peer name veth-10b
ip link set veth-01b netns mtr-test0
ip link set veth-10b netns mtr-test1

# Assign IP addresses to the interfaces in namespace mtr-test0
sudo ip netns exec mtr-test0 /bin/bash -c '
	ip addr add 172.30.1.0/31 dev veth-01a
	ip addr add 172.30.2.0/31 dev veth-01b
[...]

PR #487, opened 2023-09-30
Adjust capability handling

flu0r1ne:
This PR addresses issues related to capability handling in `mtr-packet`, as detailed in issue #485. It completely resolves Bug 1 and partially resolves Bug 2 by ensuring that `CAP_NET_RAW` and `CAP_NET_ADMIN` are retained in the permitted set. These capabilities are elevated to the effective set when necessary. This eliminates "permission denied" errors encountered when specifying a local interface using the `--interface` flag or setting a packet mark with the `-M` option. To the best of my knowledge, these issues are specific to Linux.

The commits have been tested on Linux version 6.1.51.

### DIFF
--- a/packet/construct_unix.c
+++ b/packet/construct_unix.c
@@ -32,6 +32,10 @@
 #define SOL_IP IPPROTO_IP
 #endif
 
+#ifdef HAVE_LIBCAP
+#include <sys/capability.h>
+#endif
+
 /*  A source of data for computing a checksum  */
 struct checksum_source_t {
     const void *data;
@@ -278,6 +282,67 @@ int construct_udp6_packet(
     return 0;
 }
 
+#define SET_MARK_N_ADDED_CAPS 1
+
+/* Set the socket mark */
+static int set_socket_mark(int socket, unsigned int mark) {
+    int result = -1;
+
+    // Add CAP_NET_ADMIN to the effective set if libcap is present
+#ifdef HAVE_LIBCAP
+    cap_t cap = NULL;
+    static cap_value_t cap_add[SET_MARK_N_ADDED_CAPS] = { CAP_NET_ADMIN };
+
+    // Get the capabilities of the current process
+    cap = cap_get_proc();
+    if (cap == NULL) {
+        goto cleanup_and_exit;
+    }
+
+    // Set the required capability flag
+    if (cap_set_flag(cap, CAP_EFFECTIVE, SET_MARK_N_ADDED_CAPS, cap_add,
+        CAP_SET)) {
+        goto cleanup_and_exit;
+    }
+
+    // Apply the modified capabilities to the current process
+    if (cap_set_proc(cap)) {
+        goto cleanup_and_exit;
+    }
+#endif /* ifdef HAVE_LIBPCAP */
+
+    // Set the socket mark
+    if (setsockopt(socket, SOL_SOCKET, SO_MARK, &mark, sizeof(mark))) {
+        goto cleanup_and_exit;
+    }
+
+    // Drop CAP_NET_ADMIN from the effective set if libcap is present
+#ifdef HAVE_LIBCAP
+
+    // Clear the CAP_NET_ADMIN capability flag
+    if (cap_set_flag(cap, CAP_EFFECTIVE, SET_MARK_N_ADDED_CAPS, cap_add,
+        CAP_CLEAR)) {
+        goto cleanup_and_exit;
+    }
+
+    // Apply the modified capabilities to the current process
+    if (cap_set_proc(cap)) {
+        goto cleanup_and_exit;
+    }
+#endif /* ifdef HAVE_LIBPCAP */
+
+    result = 0; // Success
+
+cleanup_and_exit:
+
+#ifdef HAVE_LIBCAP
+    cap_free(cap);
+#endif /* ifdef HAVE_LIBPCAP */
+
+    return result;
+}
+
+
 /*
     Set the socket options for an outgoing stream protocol socket based on
     the packet parameters.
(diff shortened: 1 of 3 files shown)


flu0r1ne:
I am still working on cross-platform testing this PR. Please wait until I test it further before merging . Code review would be appreciated.

rewolff:
As for "code review"... Just done that: This is the RIGHT way to do this kind of stuff: Just place a call to "just do it" in the code, and hide the ifdefs in the implementation of that call. There is way too many ifdefs in the main program flow. 
(e.g. for ipv6. I would prefer it if the main code just reads: 
```
if (IPV6_ENABLED) create_packet_ipv6 ();
else                          create_packet_ipv4 ();
``` and then the  IPV6_ENABLED is always 0 if ipv6 is not available, and returns config.ipv6_enabled if it IS available.

**Comment by rewolff** [comment]
About the cross platform testing: I created a virtualbox machine with freebsd on my workstation. Doing the same should give you the capability to test on freebsd....

**Comment by flu0r1ne** [comment]
I manually tested the changes on both FreeBSD 13.2 and Linux 6.1, and also ran them against the automated Linux tests in PR #488.

First, I tested the code on `Linux 6.1`. I specifically bound `mtr` to a device through which traffic would not usually be routed. For example, the default gateway would normally send packets to a different device. The test traced the expected network path.

```bash
./bootstrap.sh
./configure
make
MTR_PACKET="./mtr-packet" sudo -E ./mtr -t -I wlp5s0 1.1.1.1
```

Next, I configured the code on `FreeBSD 13.2`. To the best of my knowledge, `SO_BINDTODEVICE`, `SO_MARK`, and `libcap` are only supported on Linux. Therefore, FreeBSD can effectively serve as a stand-in for other non-Linux hosts, since these execution paths will be omitted.

```bash
./bootstrap.sh
./configure --without-gtk --without-jansson
make
MTR_PACKET="./mtr-packet" sudo -E ./mtr 1.1.1.1
MTR_PACKET="./mtr-packet" sudo -E ./mtr -I vtnet0 1.1.1.1
MTR_PACKET="./mtr-packet" sudo -E ./mtr -I lo1 127.0.0.1
```

The test again traced the expected path.

Finally, I validated the changes against the automated Linux tests from PR #488. The `test_interface_binding` test now succeeds.

```bash
git switch -c adjust-capability-handling-test
git merge adjust-capability-handling
git merge linux-tests
make clean
make
sudo python ./test/linux/netemtests.py
```

This confirms that the modifications are working as expected across different platforms and scenarios.

flu0r1ne:
Regarding the use of ifdefs, I concur that they can make the code more difficult to read and reason about. However, I believe most of them are necessary.

For `HAVE_LIBCAP`, which is undefined if Autoconf doesn't detect it, I can make it toggleable by incorporating an if statement within the ifdef, as follows:

```
#ifdef HAVE_LIBCAP
    if (HAVE_LIBCAP) {
        drop_excess_capabilities();
    }
#endif
```

Alternatively, to make it even more concise:

```
#if (HAVE_LIBCAP == 1)
    drop_excess_capabilities();
#endif
```

Since drop_excess_capabilities will only be used when HAVE_LIBCAP is defined, it seems reasonable to continue conditionally compiling it out when `HAVE_LIBCAP` is not present to avoid warnings.

`SO_BINDTODEVICE` and `SO_MARK` are not managed by Autoconf, so I don't believe your comment applies to these `ifdef`s.

As for improving clarity, I can introduce an enum to manage types instead of using `ifdef`s in `construct_unix.c`. The enum would define a common interface that elevates privileges on platforms with `LIBCAP` and acts as a no-op on platforms without it. I would then map these enum values to the actual capabilities in cases where the system supports them. Here is the current code:

```
/*
    This defines a common interface which elevates privileges on
    platforms with LIBCAP and acts as a NOOP on platforms without
    it.
*/
#ifdef HAVE_LIBCAP

typedef cap_value_t mayadd_cap_value_t;
#define MAYADD_CAP_NET_RAW CAP_NET_RAW
#define MAYADD_CAP_NET_ADMIN CAP_NET_ADMIN
#define MAYADD_UNUSED

#else /* ifdef HAVE_LIBCAP */

typedef int mayadd_cap_value_t;
#define MAYADD_CAP_NET_RAW ((mayadd_cap_value_t) 0)
#define MAYADD_CAP_NET_ADMIN ((mayadd_cap_value_t) 0)
#define MAYADD_UNUSED UNUSED

#endif /* ifdef HAVE_LIBCAP */
```

https://github.com/flu0r1ne/mtr/blob/04d5abe3b5d9975a4e74f15ef86eae3b355ae729/packet/construct_unix.c#L292-L306

I hope this addresses your concerns and I would appreciate any further feedback. I plan to make one additional change related to error handling. After receiving your follow-up comments, I'll make any necessary revisions, rerun the manual tests, and then the code should be ready for merging.

rewolff:
I was not criticizing you for your use of ifdefs. You were doing it mostly right in the parts that I scanned. But your example can be used as an example: 

I would like to have the main code read: 
```
  drop_excess_capabilities();  // this evaluates to a noop if we don't have LIBCAP. 
```
Then in the implementation of "drop_excess_capabilities" there would be the ifdef HAVE_LIBCAP. Or where the 

20 years ago, that would have meant a call to an empty function. 20 years ago I would've said: if it makes the code more readable, it is worth the extra/empty function call. Nowadays the compiler will remove the call during the linking stage. 

Don't optimize for run-time, optimize for readabilty.
Early nineties I worked on a project optimized for runtime: We knew in advance timing was going to be tricky. Turns out the optimized code (written before my involvement in the project) "didn't work". I rewrote for readability: All code got about 3x slower. Then I optimized just the ONE function that was taking most of the time (80% of available time after optimization). And then it worked. And within the real-time constraints that we had. 
The readability aspect of the code is WAY more important than the runtime.

PR #488, opened 2023-10-02
Linux-Only Interface, Marking, and IP Unit Tests

flu0r1ne:
This commit introduces three unit tests focused on interface binding, packet marking, and IP source address spoofing/selection. Each of these tests builds upon the original `MtrPacketTest` base class. To evaluate these network-dependent features, the tests utilize an emulated network environment. To enable this, a small network emulation library, `netem`, has been developed specifically for this purpose. `netem` allows for the creation of arbitrary network configurations for testing and can reliably set up and clean up virtual network environments on Linux systems. The only dependencies are `iproute2` and `libc`, which are generally pre-installed on most Linux hosts.

The commit adds three tests that are expected to fail (see #485):

1. Interface Binding: Evaluates the capability of `mtr-packet` to bind to a specific interface.
2. Packet Marking: Assesses the ability of `mtr-packet` to apply a Linux networking mark (fwmark).
3. Source Address Selection: Tests `mtr-packet`'s ability to spoof or bind to a specified source address.

### DIFF
--- a/test/linux/netem.py
+++ b/test/linux/netem.py
@@ -1,0 +1,667 @@
+'''
+MtrNetEm - a small network emulation library
+
+Description
+-----------
+This small, self-contained Python library serves as a high-level API for the
+creation and management of virtual network topologies in a Linux environment.
+By leveraging Linux's networking capabilities, it allows for the dynamic
+establishment of virutal network environments, links, and complex route and rule
+configurations. It only relies on iproute2, libc, and Python 3.10. The core
+architecture revolves around the `Network` base class, from which custom network
+topologies can be designed. A typical use-case involves inheriting from this base
+class and defining `Host` and `Link` objects as class attributes within the
+constructor. This design is essential because the assignment of these objects to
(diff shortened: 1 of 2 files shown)


rewolff:
The netem seems to fail LINT checks. Please look into that. 

You should've said: They "expected to fail" issues are issues that should be in progress of being fixed, but these tests actually detect that the issue is still present.

flu0r1ne:
I will investigate the linting issues. Thank you for bringing them to my attention.

You are correct, I intended for them to fail with the `main` branch to demonstrate the issues. Fixes for some of these issues are currently underway.

rewolff:
OK. great. 
I think that now that I've approved running of the tests once for your pull requests, in the future they will run automatically, so that you don't have to wait for me to be awake.

flu0r1ne:
I believe I've resolved the linting issues; my local version of flake8 is no longer reporting any errors. However, it seems that the CI  pipeline is not triggered when I push changes.

PR #493, opened 2023-10-07
Add WSL method to Windows Install

slord399:
It would be much simpler and not experience IPv6 related build fail at all.

Note: will give popup window to show progress.


rewolff:
This WSL stuff does not belong under the header "building for Windows". You should add something like: "using mtr on windows"... and then tell about WSL. 
In the building for windows section, there might be suff about building on WSL too ?

slord399:
Sure, let me edit bit to correct it.

slord399 (commented):
OK, patched

Issue #465, opened 2023-02-18
cygwin 3.4.6 failure starting overlapped thread pipe read (code 232)


userdocs:
Some recent updates of Cygwin have created this runtime error

```
failure starting overlapped thread pipe read (code 232)
mtr: Failure to start mtr-packet: Invalid argument
```

It worked on 3.3.5 before updating (possibly related to this? - https://cygwin.com/pipermail/cygwin-announce/2022-November/010810.html

but right now it's not working at all and i don't know why only guess it's x64 related or some other cygwin change.

apschultz:
The named pipe issue is odd. In my environment, it is caused by ERROR_NO_DATA which should only occur if the pipe is configured as non blocking. I added a check for ERROR_NO_DATA and a small sleep to avoid excessive CPU which effectively converts the IO thread to polling mode, but it fixes the issue.

The other build issues are also pretty easy to resolve. The w32api finally includes the ICMP6 definitions which is where most of the frustration comes from. 

`diff --git a/packet/probe_cygwin.c b/packet/probe_cygwin.c
index f41e514..fe6bd34 100644
--- a/packet/probe_cygwin.c
+++ b/packet/probe_cygwin.c
@@ -281,7 +281,7 @@ void WINAPI on_icmp_reply(
             remote_addr6->sin6_family = AF_INET6;
             remote_addr6->sin6_port = 0;
             remote_addr6->sin6_flowinfo = 0;
-            memcpy(&remote_addr6->sin6_addr, reply6->AddressBits,
+            memcpy(&remote_addr6->sin6_addr, &reply6->Address.sin6_addr,
                    sizeof(struct in6_addr));
             remote_addr6->sin6_scope_id = 0;
         }
@@ -521,7 +521,10 @@ DWORD WINAPI icmp_service_thread(LPVOID param) {

             if (!success) {
                 err = GetLastError();
-
+                               if (err == ERROR_NO_DATA) {
+                                       usleep(10);
+                                       continue;
+                               }
                 if (err != ERROR_IO_PENDING) {
                     error_win(
                         EXIT_FAILURE, err,
diff --git a/packet/probe_cygwin.h b/packet/probe_cygwin.h
index eec001f..acaec64 100644
--- a/packet/probe_cygwin.h
+++ b/packet/probe_cygwin.h
@@ -19,32 +19,14 @@
 #ifndef PROBE_CYGWIN_H
 #define PROBE_CYGWIN_H

+#include <cygwin/in6.h>
[...]

PR #495, opened 2023-10-23
Update Cygwin ICMP service thread for asynchronous pipes

matt-kimball:
Recent versions of Cygwin implement pipe() using Windows' named pipes, and put the read end of the pipe in FILE_PIPE_COMPLETE_OPERATION mode, which doesn't allow overlapped I/O operations.

For the relevant commit in the Cygwin repository, see 9e4d308cd592fe383dec58ea6523c1b436888ef8

Happily, this mode is approximately equivalent to Unix's non-blocking read mode, so we can avoid overlapped I/O entirely, and instead wait on the Windows handle for the read end of the pipe when we need an alertable wait to receive ICMP probe completions.

Thanks to Adam Schultz for research into this issue and a first attempt at a fix.

This fixes issue #465

### DIFF
--- a/packet/probe_cygwin.c
+++ b/packet/probe_cygwin.c
@@ -136,8 +136,14 @@ void init_net_state(
     net_state->platform.thread_out_pipe_read = out_pipe[0];
     net_state->platform.thread_out_pipe_write = out_pipe[1];
 
-    net_state->platform.thread_in_pipe_read_handle =
-        (HANDLE)get_osfhandle(in_pipe[0]);
+    InitializeCriticalSection(&net_state->platform.pending_request_cs);
+    net_state->platform.pending_request_count = 0;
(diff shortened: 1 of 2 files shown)


matt-kimball:
Actually, upon further investigation, WaitForSingleObject on the named pipe handle doesn't accomplish what I thought it would.  Do not merge this PR yet - I will update with a better solution.

matt-kimball:
PR updated.  Unfortunately, the solution was slightly more complicated and uses some new Windows synchronization primitives, but now actually works to block in the alertable wait.  I have tested this with newer versions of Cygwin, and it works as expected, but I also believe it will work for the older versions of Cygwin, and is less fragile since it no longer retrieves the Windows `HANDLE` of the pipe created by Cygwin and therefore doesn't depend on how Cygwin implements it.

PR #504, opened 2024-03-04
Handle EHOSTDOWN errno.

arekm:
Handle EHOSTDOWN which occurs on MacOS and causes:

% mtr 192.168.68.2
mtr: Unexpected mtr-packet error

for unreachable host.

### DIFF
--- a/packet/probe_cygwin.c
+++ b/packet/probe_cygwin.c
@@ -136,8 +136,14 @@ void init_net_state(
     net_state->platform.thread_out_pipe_read = out_pipe[0];
     net_state->platform.thread_out_pipe_write = out_pipe[1];
 
-    net_state->platform.thread_in_pipe_read_handle =
-        (HANDLE)get_osfhandle(in_pipe[0]);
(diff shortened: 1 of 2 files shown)


rewolff:
The code snippet around you seems to collapse multiple error codes down to just one Your added line "joins the crowd". 

I personally strongly dislike this collapsing of error codes. It would be very much appreciated if you figure out where the further processing of these error messages from mtr-packet happen and then add new error codes there as well...

arekm:
Did that way because existing code does it already

````
 } else if (errno == ENETUNREACH) {
        printf("%d no-route\n", command_token);
    } else if (errno == EHOSTUNREACH) {
        printf("%d no-route\n", command_token);
````
and EHOSTDOWN fits in that.

These would also need to be split.

rewolff:
Correct! 
So since you're already "working on that code" I'd appreciate the help if you could "improve" it further by splitting it now. Maybe add a comment that I don't want error codes collapsed in the future.

PR #513, opened 2024-08-21
Handle EHOSTDOWN and refine error handling better granularity

arekm:
Handle EHOSTDOWN which occurs on MacOS and causes:

% mtr 192.168.68.2
mtr: Unexpected mtr-packet error

for unreachable host and also split error handling better granularity.

### DIFF
--- a/man/mtr-packet.8.in
+++ b/man/mtr-packet.8.in
@@ -317,13 +317,17 @@ label, and so on.  The values are provided in this order:
 .IR ttl .
 .HP 7
 .TP
-.B no-route
-There was no route to the host used in a
+.B no-route-network
+.B no-route-host
+There was no route to the network or the host itself for the
 .B send-probe
-request.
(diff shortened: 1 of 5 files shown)


arekm:
That's continuation of https://github.com/traviscross/mtr/pull/504

arekm:
There is one thing there:
````
} else if (icmp_type == ICMP_DEST_UNREACH) {
-        result = "no-route";
+        /* XXX icmphdr->code is not known here, so assume that host is unreachable */
+        result = "no-route-host";
````

accessing icmphdr->code for ipv4 is possible but requires a bit of changes and not sure if it's worth the hassle.

Also ipv6 has no host vs network unreachable differentiation as I see it:
https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xhtml#icmpv6-parameters-3

Issue #522, opened 2025-02-14
Leading spaces in the XML output


marek22k:
Hello,

I am currently trying out the XML output function and I have noticed that for Loss, Snt, Last, ... have several leading spaces. If you want to parse the output and display it in a GUI or similar, you would have to remove the spaces first. Therefore, in my opinion, it would make sense to output it without spaces.

```
$ mtr --report -c 3 --xml 1.1.1.1
<?xml version="1.0"?>
<MTR SRC="p2pnode" DST="1.1.1.1" TOS="0x0" PSIZE="64" BITPATTERN="0x00" TESTS="3">
    <HUB COUNT="1" HOST="gw01.netcup.net">
        <Loss>  0.0%</Loss>
        <Snt>     3</Snt>
        <Last>   0.5</Last>
        <Avg>   2.2</Avg>
        <Best>   0.5</Best>
        <Wrst>   5.3</Wrst>
        <StDev>   2.6</StDev>
    </HUB>
    <HUB COUNT="2" HOST="94.16.25.158">
        <Loss>  0.0%</Loss>
        <Snt>     3</Snt>
        <Last>   0.4</Last>
        <Avg>   0.4</Avg>
        <Best>   0.4</Best>
        <Wrst>   0.4</Wrst>
        <StDev>   0.0</StDev>
    </HUB>
    <HUB COUNT="3" HOST="ae0-0.bbr01.anx84.nue.de.anexia-it.net">
        <Loss>  0.0%</Loss>
        <Snt>     3</Snt>
        <Last>   4.0</Last>
        <Avg>   4.1</Avg>
        <Best>   4.0</Best>
        <Wrst>   4.1</Wrst>
        <StDev>   0.0</StDev>
    </HUB>
    <HUB COUNT="4" HOST="ae2-0.bbr02.anx25.fra.de.anexia-it.net">
        <Loss>  0.0%</Loss>
        <Snt>     3</Snt>
        <Last>   4.1</Last>
        <Avg>   4.1</Avg>
        <Best>   4.1</Best>
        <Wrst>   4.1</Wrst>
        <StDev>   0.0</StDev>
    </HUB>
    <HUB COUNT="5" HOST="ae1-0.bbr02.anx82.fra.de.anexia-it.net">
        <Loss>  0.0%</Loss>
        <Snt>     3</Snt>
        <Last>   4.8</Last>
        <Avg>   9.3</Avg>
        <Best>   4.0</Best>
[...]

PR #524, opened 2025-02-14
Fix 522

marek22k:
Fix https://github.com/traviscross/mtr/issues/522

### DIFF
--- a/ui/report.c
+++ b/ui/report.c
@@ -406,6 +406,7 @@ void xml_close(
     int i, j, at, max;
     ip_t *addr;
     char name[MAX_FORMAT_STR];
+    char buf[128];
 
     printf("<?xml version=\"1.0\"?>\n");
     printf("<MTR SRC=\"%s\" DST=\"%s\"", ctl->LocalHostname,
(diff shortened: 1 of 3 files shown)


rewolff:
I replied by Email. 

After three such replies I decided to check if they came through.... I hit refresh for a couple of minutes, but nothing and then got distracted. So this morning I checked and my reply is not here. :-(

> I also find it a bit strange that the capitalization in the XML
> output is not uniform. `<?xml` is written in lower case. Things like
> `MTR` and `HUB` are capitalized and for `Loss` and so on only the
> first letter is capitalized.

xml is part of the language. Needs to be lower case. 

Further, MTR and HUB are capitalized. A bit like a chapter heading
being larger.

The others are "info fields, and my persional taste is not to
captialize those. Someone who wtote the XML seems to have thought
otherwise.

As people by now will have written programs to parse the XML, I don't
think we could change those. 

About the spaces: I agree. And I think we could remove them. Will
probably cause some headaches, but those are caused by bad programming
practices, so tolerable that we break things. 

So.. You're invited to fix the spaces issue and submit a pull request.

marek22k:
So you don't want to write the tags all in lower case because of backwards compatibility? Couldn't you just say: New mtr version, new output format?

rewolff:
> Backward compatibility. 
Right. 

For someone whose job it is to say "maintain the mailserver", then an upgrade that causes some script or homemade addon to not work.... Hopefully he still knows how the script works, and with week of work, he can get it working again. 

For me who has to do the mailserver maintenance as a side-job such an "upgrade" that breaks compatibility would be impossible to handle. I can't take a week off my other work to fix that mailserver script that I wrote 5 years ago. 

Yes, "all lower case"  looks nicer to me and you, but I don't think the hassle for those who built on the XML output format are going to appreciate the "looks nicer" change.

rewolff:
I don't think we can change the tags. XML is meant to be
computer-parsed and that's what it is being used for. Please
remove the "fixing" of the tags.

Issue #475, opened 2023-05-23
--psize option to randomize packet size


yvs2014:
there's some error in signed/unsigned integer comparison which blocks packetsize randomizing:
```
$ sudo ./mtr --psize=-100 localhost
./mtr: invalid argument: '-100'
```
possible fix
```
--- a/ui/utils.c
+++ b/ui/utils.c
@@ -87 +87 @@ int strtonum_or_err(
-                if (num < INT_MAX)
+                if (((long)num) < INT_MAX)
```
https://github.com/yvs2014/mtr085/commit/c6a53de814bd8030381e9b7e30f6799ddf4963f4

then, it's good to limit possible values of 'psize' because
```
$ sudo ./mtr --psize=-28 localhost
floating point exception  sudo ./mtr --psize=-28 localhost
```
possible fix with
```
+            if (abs(ctl->cpacketsize) <= MINPACKET) {
+                ctl->cpacketsize = (ctl->cpacketsize > 0) ? (MINPACKET + 1) : -(MINPACKET + 1);
+            } else if (abs(ctl->cpacketsize) > MAXPACKET) {
+                ctl->cpacketsize = (ctl->cpacketsize > 0) ? MAXPACKET : -MAXPACKET;
+            }
```
https://github.com/yvs2014/mtr085/commit/34ea504a3bb2acc50560063dece653ee88e912b4

marek22k:
Related: https://github.com/traviscross/mtr/issues/523

marek22k:
Part of this issue was successfully resolved in https://github.com/traviscross/mtr/pull/525. It is now possible to specify negative (and therefore random) values. A check should now be carried out to ensure that the values are not too small.

rewolff:
IMHO that should be implemented as "random packet size between the
minimum and provided maximum". This should be done at a low level,
because min packet sizes might differ between requested protocols.

A slightly different implementation might be to consider the provided
(or random) packet size as a lower limit and the packet sending
function will pad to that size, if necessary. That will cause a much
larger percentage of packets being exactly the min packet size.

In the first case you specify the amount of padding, but for
convenience you specify the whole packet size. (consider that a "units
conversion"). In the second case, the implmentation/protocol dictates
the minimum and that is optionally padded to the provided size.

marek22k:
https://github.com/traviscross/mtr/pull/527 tries to solve the problem.
mtr seems to automatically make appropriate adjustments if the package size is too small.
[...]

PR #527, opened 2025-02-15
Fix https://github.com/traviscross/mtr/issues/475

marek22k:
With an input of `-28` the expression `(-ctl->cpacketsize - MINPACKET)` became 0, which triggered a zero division. This error is fixed by introducing a check to see if there is any room for randomness. Furthermore, a check of the input arguments in the command line and in curses is performed.
MINPACKET assumes IPv4, so these checks do not apply to IPv6. mtr still seems to work with IPv6, so it is likely that there is an additional check elsewhere.
I think it's still okay to assume IPv4 for MINPACKET (the minimum size of a packet), since IPv4 packets are smaller than IPv6 packets (due to the IP header).

### DIFF
--- a/ui/curses.c
+++ b/ui/curses.c
@@ -202,7 +202,10 @@ int mtr_curses_keyaction(
             buf[i++] = c;       /* need more checking on 'c' */
         }
         buf[i] = '\0';
-        ctl->cpacketsize = atoi(buf);
+        int new_packetsize = atoi(buf);
+        if (abs(ctl->cpacketsize) >= MINPACKET && abs(ctl->cpacketsize) < MAXPACKET) {
+            ctl->cpacketsize = new_packetsize;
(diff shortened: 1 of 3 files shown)


marek22k:
@yvs2014 What do you think of this solution? With this solution, ICMP packets with a payload of 0 bytes can still be generated.

yvs2014:
If there's no zero division, I think it's better.

> ICMP packets with a payload of 0 bytes can still be generated.
>
Frankly speaking I forgot test cases used at the time of fixing (it was for old one-thread mtr fork).
It's not connected directly with packet size randomness, but it's not totally clear at which level the payload is considered here. For example setting psize=28 I got 20 byte ICMP payload, so payload level needs to be clarified. Probably with @rewolff help we can clarify this point.

PR #530, opened 2025-04-09
mtr braille graph support

bartman:
This PR adds support for `--displaymode 3` which renders a graph using unicode braille characters.

image

Changes include:
* modify autoconf to detect `libncursesw` for wide char support
* only build `WITH_BRAILLE_DISPLAY` when wide support is detected
* since braille chars can show 2x the information, increases `SAVED_PINGS` to 400
* added `DisplayModeBraille` which allows setting `--displaymode 3`
* all rendering changes are in `ui/curses.c`
* with `DisplayModeBraille` selected, `mtr_curses_graph` will call `mtr_fill_graph_braille` instead of `mtr_fill_graph`
* braille characters can render 4 levels of dots, but I found that I could only use 3 to avoid lines bleeding into one another, while the bottom row becomes the origin line
* for each host, the values collected are scaled to fit the 3 dots of precision provided by the medium
* I've also used the `🮐` character to indicate maximum value observed -- I have mixed feeling about how it looks, but it does stand out.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -112,6 +112,20 @@ AS_IF([test "x$with_jansson" = "xyes"],
     [with_jansson=no])
 ])
 
+# Find ncursesw
+AC_ARG_WITH([ncursesw],
+  [AS_HELP_STRING([--without-ncursesw], [Build without the ncursesw interface])],
+  [], [with_ncursesw=yes])
+AS_IF([test "x$with_ncursesw" = "xyes"],
+
+  # Prefer ncursesw, if available
+  [AC_SEARCH_LIBS(
+    [wprintw], [ncursesw],
+    [AC_DEFINE([HAVE_CURSESW], [1], [Define if a ncursesw library available])],
+    [with_ncursesw=no])
+])
+AM_CONDITIONAL([WITH_CURSESW], [test "x$with_ncursesw" = xyes])
+
 # Find ncurses
 AC_ARG_WITH([ncurses],
   [AS_HELP_STRING([--without-ncurses], [Build without the ncurses interface])],
@@ -121,7 +135,7 @@ AS_IF([test "x$with_ncurses" = "xyes"],
   # Prefer ncurses over curses, if both are available.
   # (On Solaris 11.3, ncurses builds and links for us, but curses does not.)
   [AC_SEARCH_LIBS(
-    [initscr], [ncurses curses],
+    [initscr], [ncursesw ncurses curses],
     [AC_DEFINE([HAVE_CURSES], [1], [Define if a curses library available])],
     [with_ncurses=no])
 ])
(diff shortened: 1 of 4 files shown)


rewolff:
I'm a bit wary about the "set locale="" '  that you mention. If people have SOMETHING in their local language, wouldn't that mess with things? e.g. mtr needs to display a "host unreachable" message, that would normally be translated by perror, and the local characters used would work just fine... but now???

I'm afraid for such regressions that are difficult for you-and-me to test.

bartman:
> I'm a bit wary about the "set locale="" ' that you mention.

Let me see if I can figure out another way to get around it.

bartman:
@rewolff - I changed the code to use `printw` instead of `wprintw`, and was able to remove the forcing of `LOCALE=""`.  To be honest, I still don't understand why it didn't work before.

I cleaned up a bit, did a clean build, and retested on:
- gnome-terminal
- uxterm
- kitty
- ghostty

All seems to work now... but I'll leave it in "draft" until I do a bit more testing.

I will try to build this w/o ncurses... just to make sure I didn't mess up the automake/autoconfig setup.

Let me know if you want me to do any further testing, or code cleanup, or whatever.

rewolff:
I didn't yet have time to do "code review by a human" yet. I just reacted to your description, which seems to have sorted some effect. :-) 
So.... To optimize my time... can I postpone my review until you're done with those tests?

bartman:
I built a tiny debian docker for testing.  Built with and without wide support, using `configure --without-ncursesw`.  Ran again in kitty/uxterm/ghostty/gnome-terminal.  Seems to be functional w/o any LOCALE overrides.

image

I'm going to flip this back to "ready for review".

configure.ac
```diff
@@ -153,6 +167,15 @@ AS_IF([test "x$WANTS_IPV6" = "xyes"], [
...
 ])
 
+AC_ARG_ENABLE([braille],
+  [AS_HELP_STRING([--disable-braille], [Do not enable barille character graphs])],
+  [WANTS_BRAILLE=$enableval], [WANTS_BRAILLE=yes])
+
+AS_IF([test "x$WANTS_BRAILLE" = "xyes"], [
+  AC_DEFINE([ENABLE_BRAILLE], [1], [Define to enable IPv6])
+  USES_BRAILLE=yes
+])
```
rewolff, line 177: This needs a small correction in the help text.

ui/curses.c
```diff
@@ -810,6 +986,11 @@ void mtr_curses_open(
     int bg_col = 0;
     int i;
 
+#ifdef WITH_BRAILLE_DISPLAY
+    // initialize all locale variables, before ncurses starts
+    setlocale(LC_ALL, "");
+#endif
```
rewolff, line 992: You said you had removed this...

bartman: I removed `setenv("LOCALE", "", 1)` which was the hack.

If you look at the man page for locale you'll find...
> The argument category determines which parts of
> the program's current locale should be modified.
>
> LC_ALL                  - All of the locale

and
> If locale is an empty string, "", each part of the locale that should be
> modified is set according to the environment variables.

This is the way one enables locale support, using environment variables set by the user.

configure.ac
bartman: fixing.

bartman:
thanks for accepting the patch.  I look forward to seeing it in upstream some day.

yvs2014:
> setlocale()

Locales can be a bit tricky, there's easy to find out inconsistency with decimal delimiters (like inconsistency in setting interval `-i0.1` (via cli) versus `0,1` (in curses mode)) but there could be other possible side-effects connected with libc functions depended on locale.

bartman:
@yvs2014

Since argument parsing happens before `setlocale`, there should be no change to argument parsing.

I'm not aware of any input taken after `initscr`, am I missing something?

Can you provide something I can test that fails?  Even if it's a hunch... I'm willing to dig deeper myself.

yvs2014:
> I'm not aware of any input taken after initscr
>
   for example press 'i' for input (with any locale that has decimal-point distinguished from '.')
   
> test that fails
>
  I'd not say it fails, in case of input it just needs to be in accordance with a current locale (if LC_ALL is set curses-globally).
  
  Speaking more generally about locales, what I'd suppose to see that localized input/output corresponds each other.

PR #531, opened 2025-04-13
Set UTF-8 encoding for XML reports

marek22k:
I think it is a good idea to set the encoding of XML documents to UTF-8, so that a consistent result can be achieved on every platform. Furthermore, the document (can) exists without an external DTD.

### DIFF
--- a/ui/mtr.c
+++ b/ui/mtr.c
@@ -44,6 +44,7 @@
 #include <assert.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <locale.h>
 #include <sys/stat.h>
 #include <sys/time.h>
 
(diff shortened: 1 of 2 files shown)


marek22k:
@rewolff Can you review the PR?

rewolff:
I strongly hope that if I manually parse this, I will not encounter "weird characters" . i.e. that backwards compoatibility is maintained. Maybe if you have "weird" international names on the route, you'd see any change. 

Right?

marek22k:
Mhh, actually UTF-8 should be able to represent "strange characters" (and is therefore "better" than others).
The assumption that the user has always set a suitable charset for all characters is certainly not always correct.
If we do not use UTF-8, XML parsers will try to "guess" the charset. UTF-8 is, as far as I can see, standard for many XML documents.

marek22k:
I've just been thinking again: Can an XML report contain special characters at all? DNS names cannot contain arbitrary characters. From this point of view, there is only the danger that normal characters are not supported due to an incorrect charset (set by the user) (which would be solved by UTF-8).

Issue #533, opened 2025-06-06
BUG: Segmentation Fault when trying to diagnose '???' host


RPN17M2G:
Found by Omer Gindi - RPN17M2G

Enviourment:
Version: mtr 0.95
OS:  Ubuntu 24.04.2 LTS on WSL
WSL:
    WSL version: 2.5.7.0
    Kernel version: 6.6.87.1-1
    WSLg version: 1.0.66
    MSRDC version: 1.2.6074  
    Direct3D version: 1.611.1-81528511
    DXCore version: 10.0.26100.1-240331-1435.ge-release
    Windows version: 10.0.26100.4061

Starting using the tool - everything is fine:

![Image](https://github.com/user-attachments/assets/ad41ee94-6e41-4aa1-a600-a8f4972a0ccb)

Then I wanted to see what is the "???" host so I clicked on "set as new destination":

![Image](https://github.com/user-attachments/assets/c72de231-0e42-4648-917c-9e4cf3a4eec9)

Then the window got stuck and the cli filled with:
"mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::"

Screenshot of the stuck window - unstopped printing of the "couldn't find host...":

![Image](https://github.com/user-attachments/assets/749d91fe-bd04-41cb-a809-c8f0344785e2)

Until segmentation fault - after clicking abort on the UI

...
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
mtr: dns_ack: Couldn't find host ::
Segmentation fault (core dumped)

rewolff:
Three issues here. 
1) It doesn't make sense to click on the ??? Don't click there. 
2) The ???? should not be click-able. 
3) mtr shouldn't crash with a not-resolvable name. 

Re: 1: Lets consider an analogy. Normally when you send a parcel to someone at the other side of the country, you pay once and they check that once and everything is fine. But let's change the system: each sorting station along the way needs to be paid one coin and you need to put enough coins on the outside  of the parcel to get it to the destination. 

[...]

PR #534, opened 2025-06-14
net: implement addrcmp for AF_UNSPEC

jlu5:
When mtr is compiled with IPv6 enabled, the GTK frontend sets the address family to AF_UNSPEC when changing destinations. However, addrcomp does not support this AF type, which causes all comparisons to not match. For the GTK frontend, this leads to a cascading failure when setting the destination to a host that does not resolve:

1. Upon setting `ctl->af = AF_UNSPEC`, `net_max()` will report that all hops have data, as no hops can match the previous remote address.
2. The GTK frontend tries to render a row for every hop, including its IP / hostname.
3. The guard in the GTK frontend to avoid looking up unknown addresses also fails, causing mtr to flood DNS packets and effectively hang.

Fixes https://github.com/traviscross/mtr/issues/533

### DIFF
--- a/ui/net.c
+++ b/ui/net.c
@@ -895,6 +895,7 @@ int addrcmp(
         break;
 #ifdef ENABLE_IPV6
     case AF_INET6:
+    case AF_UNSPEC:
         rc = memcmp(a, b, sizeof(struct in6_addr));
         break;
 #endif


rewolff:
What happens if there is no IPV6 support? Is the AF also set to AF_UNSPEC? What if IPV6 is enabled, but the current invocation is in IPV4 mode? Won't the compare then fail in the old way?

rewolff:
OK. Lets just hope the one liner doesn't break anything. :-)

yvs2014:
> not sure it's the best idea to potentially compare an in_addr struct with an in6_addr
>
  with af=0 there might be side effects in other parts too (not only in addrcmp()) which hard to detect,
  for example one on the surface: run with ipv4 target, then set dest from a row with "???", and then unpause:
  "invalid remote IP address: Address family not supported by protocol" from https://github.com/traviscross/mtr/blob/master/ui/cmdpipe.c#L359 trying to resolve ipv4 target address with af=0.
  
  What about asserting 'address family' to catch these cases in gtk code?
```  
--- a/ui/net.c
+++ b/ui/net.c
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <sys/select.h>
 #include <unistd.h>
+#include <assert.h>
 
 #ifdef HAVE_ERROR_H
 #include <error.h>
@@ -895,10 +896,11 @@ int addrcmp(
         break;
 #ifdef ENABLE_IPV6
     case AF_INET6:
-    case AF_UNSPEC:
         rc = memcmp(a, b, sizeof(struct in6_addr));
         break;
 #endif
+    default:
+        assert(family);
     }
 
     return rc;
```

yvs2014:
In case of setting a new destination in GUI version, probably some cases could be avoided by testing if it's resolvable,
something like
```c
--- a/ui/gtk.c
+++ b/ui/gtk.c
@@ -728,6 +728,18 @@ void gtk_loop(
     gtk_main();
 }
 
+static int IsDestResolvable(struct mtr_ctl *ctl, const gchar *hostname) {
+    struct addrinfo *res = NULL;
+    int af = ctl->af;
+    ctl->af = DEFAULT_AF;
+    int rc = get_addrinfo_from_name(ctl, &res, hostname) == 0;
+    if (!rc)
+        ctl->af = af;
+    if (res)
+        freeaddrinfo(res);
+    return rc;
+}
+
 static gboolean NewDestination_activate(
     GtkWidget * widget ATTRIBUTE_UNUSED,
     gpointer data)
@@ -737,7 +749,7 @@ static gboolean NewDestination_activate(
     GtkTreePath *path = (GtkTreePath *) ctl->gtk_data;
 
     hostname = getSelectedHost(path);
-    if (hostname) {
+    if (hostname && IsDestResolvable(ctl, hostname)) {
         ctl->gtk_data = hostname;
         gtk_entry_set_text(GTK_ENTRY(Entry), hostname);
         Host_activate(Entry, ctl);
```
but sure it needs to be tested (I'm not too much familiar with gtk3 code in mtr)

PR #569, opened 2026-05-01
ui: throttle curses redraws during packet bursts

Komzpa:
## Summary

Throttle interactive curses redraws in the main `select()` loop to at most once every 100 ms.

Packet, resolver, and keyboard file descriptors are still serviced immediately, but terminal repainting is scheduled separately and folded into the existing `select()` timeout. This targets profiles where packet/DNS bursts drive repeated ncurses `wnoutrefresh()` / `werase()` work faster than a human-visible terminal needs.

## Validation

- `git diff --check`
- `./bootstrap.sh && ./configure && make -j2`
- `make check` was attempted. `test/cmdparse.py` passed, but `test/param.py` failed because the local unprivileged test runner could not open the packet listening socket (`Operation not permitted`), and `test/probe.py` failed on network/probe expectations such as `no-route-host` / `no-reply` instead of `ttl-expired`.

### DIFF
--- a/ui/select.c
+++ b/ui/select.c
@@ -41,6 +41,56 @@
 #include "display.h"
 #include "select.h"
 
+#define MIN_DISPLAY_REDRAW_USEC 100000
+
+static int timeval_after_or_equal(
+    const struct timeval *a,
+    const struct timeval *b)
+{
+    if (a->tv_sec > b->tv_sec)
+        return 1;
+    if (a->tv_sec < b->tv_sec)
+        return 0;
+
+    return a->tv_usec >= b->tv_usec;
+}
+
+static void timeval_add_usec(
+    struct timeval *timeval,
+    int usec)
+{
+    timeval->tv_usec += usec;
+    timeval->tv_sec += timeval->tv_usec / 1000000;
+    timeval->tv_usec %= 1000000;
+}
+
+static void timeval_subtract(
+    struct timeval *result,
+    const struct timeval *end,
+    const struct timeval *start)
+{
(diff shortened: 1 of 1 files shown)


rewolff:
Thanks for the patch. 
Question: If I trace a host say 5 hops away with an interval of say 30 second, I would send out a packet about every 6 seconds. Wouldn't the curses redraw only update every 6 seconds? 

The long intervals are used for long running monitoring of slow-ish links.

rewolff:
OK. I apologize: I wrote that before checking the code. My objection, you've already implemented, right? Please confirm I'm reading the code correctly.

Komzpa:
Yes, that's how I intended it to work.

The redraw itself is capped by `MIN_DISPLAY_REDRAW_USEC`, but the `select()` timeout is also limited to the next redraw deadline:

```c
timeval_subtract(&redrawtime, &nextredraw, &thistime);
timeval_min(&selecttime, &redrawtime);
```

So packet/resolver events are still handled immediately, and when the probe interval is long the curses UI can still wake up for a redraw before the next packet is sent. The throttle only prevents repainting on every packet burst event.

rewolff:
Internet is down here today. 

OK. Thanks for the confirmation. Remind me if I didn't already 
merge it. I can't (easily) check if its already merged today. 
I'm considering it handled, remind me with a short reaction if
it isn't already merged, ok?

Komzpa:
It's not yet merged by you. Helloes from Batumi where internets are also tricky :)

PR #570, opened 2026-05-07
ci: fix GitHub Actions test workflow

Komzpa:
## Summary

Fix the current GitHub Actions failures on `master`:

- install `pbr==7.0.3` with the pinned flake8/bandit toolchain so `flake8-bandit` can import `bandit`, which imports `pbr.version`
- run `test/cmdparse.py` through `sudo` because it launches `mtr-packet`, and `mtr-packet` needs raw socket permissions for the exercised command parser paths

## Validation

- `uv run --python 3.9 --with flake8==3.9.2 --with flake8-bandit==2.1.2 --with bandit==1.7.2 --with pbr==7.0.3 python -m flake8 .`
- `./bootstrap.sh && ./configure --without-gtk --without-jansson && make -j $(nproc)`
- `./mtr --report --report-cycles 1 -m 1 1.1.1.1`
- `sudo python3 ./test/cmdparse.py` (`6 tests OK`)

The latest upstream failure I checked was `Test compilation` run `25505515657`, where lint failed with `ModuleNotFoundError: No module named 'pbr'` and `cmdparse.py` timed out after `mtr-packet` failed to open IPv4/IPv6 sockets with `Permission denied`.

PR #571, opened 2026-05-07
packet: document and constrain default UDP ports

### LINKED ISSUE #517: [DOC] Please, document which destination ports are used by default when using -u or -T option
...and the reasoning of this. Thanks!

P.S: I've just found https://unix.stackexchange.com/questions/452709/what-udp-destination-ports-does-mtr-u-use

Komzpa:
## Summary

Document the default port behavior requested in #517 and address the follow-up from @rewolff:

- TCP and SCTP traces use target port 80 when `--port` is not given
- UDP traces use the probe sequence number as the destination port when `--port` is not given
- UDP source port defaults to the low 16 bits of the mtr process ID, adjusted out of the privileged port range if needed
- when `--port` is set for UDP, the sequence number moves to the source port as before

This matches `ui/mtr.c` and `packet/construct_unix.c`, and avoids defaulting UDP probes to a privileged source port when modern large PIDs truncate below 1024.

Fixes #517.

## Validation

- `make mtr.8`
- `./bootstrap.sh && ./configure --without-gtk --without-jansson && make -j $(nproc)`
- `./mtr --report --report-cycles 1 -m 1 1.1.1.1`
- `sudo python3 ./test/cmdparse.py` (`6 tests OK`)
- `git diff --check`

## CI note

This is intentionally left as a draft until #570 lands. The current failing checks on this branch are the existing `master` workflow failures fixed by #570, not failures introduced by this branch.

### DIFF
--- a/man/mtr.8.in
+++ b/man/mtr.8.in
@@ -446,9 +446,16 @@ Use Stream Control Transmission Protocol packets instead of ICMP ECHO.
 .TP
 .B \-P \fIPORT\fR, \fB\-\-port \fIPORT
 The target port number for TCP/SCTP/UDP traces.
+When this option is not given, TCP and SCTP traces use port 80.
+UDP traces encode the probe sequence number in the destination port
+instead, so the destination port changes for each probe.
 .TP
 .B \-L \fILOCALPORT\fR, \fB\-\-localport \fILOCALPORT
-The source port number for UDP traces.
+The source port number for UDP traces.  When this option is not given,
+UDP traces use the mtr process ID as the source port, unless
+.B \-\-port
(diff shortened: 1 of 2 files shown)


rewolff:
Hi. 
Can you do me a favor? The "use the process number as the port" was reasonable when process numbers used to be 15 bits. 

Nowadays pids are probably 31 or 32 bits. So the assignment needs to be modified to: 
- truncate the PID to 16 bits. 
- stay away from privileged ports (by default) 
Can you "while you're at it" provide a patch for that too?

Komzpa:
@rewolff sure, here it is. If you have other favorite/annoying issues throw them at me too while my checkout is still hot, will take a look.

PR #575, opened 2026-05-07
csv: omit trailing empty header field

### LINKED ISSUE #451: Fix header line generation for CSV data export
I built the evolving network analysis tool according to the development revision “v0.95-17-g826ffa9”.
I tried the program out then accordingly.

```
Sonne:/home/altes_Heim2/elfring/Projekte/Bau/mtr/console # ./mtr --report-cycles 5 --report --csv download.opensuse.org
Mtr_Version,Start_Time,Status,Host,Hop,Ip,Loss%,Snt, ,Last,Avg,Best,Wrst,StDev,
…
```

:thought_balloon: Such a header line should not contain two empty field names (in the middle and at the end).

Komzpa:
## Summary

Stop emitting a trailing delimiter at the end of the CSV header line.

The existing spacer field behavior is preserved: if the active field list contains a space field, CSV output keeps that column so existing column positions do not move. This only removes the unintended final empty header field.

Fixes #451.

## Validation

- `./bootstrap.sh`
- `make -j32`
- `sudo -n ./mtr --report-cycles 1 --report --csv 127.0.0.1`
- Python `csv` check verifying:
  - header/data row counts match
  - the spacer column is still present as an empty CSV field
  - the last header/data field is not empty
- `git diff --check`

### DIFF
--- a/ui/report.c
+++ b/ui/report.c
@@ -570,10 +570,10 @@ void csv_close(
         snprint_hop_name(ctl, name, sizeof(name), at, addr);
 
         if (at == net_min(ctl)) {
-            printf("Mtr_Version,Start_Time,Status,Host,Hop,Ip,");
+            printf("Mtr_Version,Start_Time,Status,Host,Hop,Ip");
 #ifdef HAVE_IPINFO
             if (ipinfo_field_selected(ctl, 0)) {
-                printf("Asn,");
+                printf(",Asn");
             }
 #endif
             for (i = 0; i < MAXFLD; i++) {
@@ -584,7 +584,7 @@ void csv_close(
                     printf(",");
                     continue;
                 }
-                printf("%s,", data_fields[j].title);
+                printf(",%s", data_fields[j].title);
             }
             printf("\n");
         }
@@ -608,7 +608,6 @@ void csv_close(
                 printf(",");
                 continue;
             }
-
             /* 1000.0 is a temporary hack for stats usec to ms, impacted net_loss. */
             if (strchr(data_fields[j].format, 'f')) {
                 printf(",%.2f",
@@ -665,7 +664,6 @@ void csv_close(
                             printf(",");
                             continue;
                         }
-
                         /* 1000.0 is a temporary hack for stats usec to ms, impacted net_loss. */
                         if (strchr(data_fields[j].format, 'f')) {
                             printf(",%.2f",


rewolff:
Can you explain to me why I keep seeing the "15 files changed" and the same "patches" in the help-output, but still "no merge conflicts" ?

Komzpa:
You were right: this branch had accidentally picked up sibling stack commits. GitHub could still merge it cleanly, so it showed no merge conflict, but the PR comparison included those unrelated help-output/curses patches.

I rebuilt the branch on current `master`; it is now a single commit touching only `man/mtr.8.in` and `ui/report.c`, and the exact-head `lint` and `compile-linux` checks are green.

rewolff:
FYI: The first time I thought "hmm". The second time "oh well" and only a bit later did I start "complaining". 

Could you check the other open pull requests as well?

rewolff:
Now that the patch is "readable" for me.... 

What I think is happening is that you've eliminated the "empty fields" that you'd get when the format string contains a space. 

I'm not sure I want to do that: Why would people want to use CSV? -- Because they want to process the data further. 
So what happens when you change this behaviour? -- scripts will be looking in the wrong columns for their data. 

People who didn't want the empty columns already had the option to not use spaces in the format string. 

And when importing the CSV into a spreadsheet, having a way to create an empty column may be handy: Sometimes it provides a space for a formula that is based on the columns near it. (Sure, some users are fine with the "instructions" go to colmn 5, rightclick, select add column and proceed from there. I would not be: that can be automated: Add an empty column in the input file!) 

So, you need to convince me that this is a good thing.

Komzpa:
Thanks, checked.

There are no other currently open PRs from me in this repository besides #575. #575 is now rebuilt as one commit touching only `man/mtr.8.in` and `ui/report.c`, with `lint` and `compile-linux` green.

I also checked the same-day merged batch. Some of those did have the same stacked-branch review problem before merge. In particular, #639 included the stacked heads for #640-#645, and #636 included #638, so GitHub marked those sibling PRs merged together. #637, #580, and #576 also had noisy PR views earlier, although their final merge diffs into `master` ended up focused after prerequisites had landed.

Sorry for the confusing review surface; I’ll avoid sending stacked PR branches like that here.

Komzpa:
That makes sense as a compatibility concern.

The reason I sent this version is that in #451 you wrote that either removing the trailing comma or removing the empty header and `0` column would be acceptable:
https://github.com/traviscross/mtr/issues/451#issuecomment-1505387596

This patch implements the second option, plus the trailing-comma cleanup. But I agree your point here is important: if users intentionally use the space field to reserve a CSV column, removing it changes column positions.

Would you prefer that I narrow this PR to only remove the trailing comma, and leave explicit spacer fields intact? That would fix the unintentional final empty field without changing the existing column layout contract.

rewolff:
Yeah, I think that's best. Just remove the empty column at the end. If someone imports it into a spreadsheet, current spreadsheets won't require you to "add a column" at the end: You can just start using more and more columns as you like, right?

Sorry for "changing my mind". Sometimes the better opinion comes a bit later. after I've already publicly stated a different one. Sorry! :-)

PR #583, opened 2026-05-07
build: honor transformed mtr-packet name

### LINKED ISSUE #191: Issue with "--program-suffix" option to ./configure
The configure script has options to add prefixes and suffixes to the installed binaries, as to let people have multiple versions installed side-by-side.

However, the makefile/code does not account for the use of that option, as it references "mtr-packet", when, in reality, the binary could be installed as something else.

For example, by cloning from git, running the bootstrap, then this:
```
# ./configure --prefix=/usr/local --without-gtk --with-bashcompletiondir=/etc/bash_completion.d --program-suffix='-0.87.223-fa75'
```

`make` will run to completion without issues, as will `make install-data-am`, but `make install-exec-am` will fail due to the failure to account for the modified program names within `make exec-hook`:
```
# make -n install-exec-hook
`setcap cap_net_raw+ep /usr/local/sbin/mtr-packet` \
	|| chmod u+s /usr/local/sbin/mtr-packet
```
The actual name of the binary in this case would be `/usr/local/sbin/mtr-packet-0.87.223-fa75`.

beyond that, `execute_packet_child()` within [ui/cmdpipe.c](https://github.com/traviscross/mtr/blob/master/ui/cmdpipe.c#L212) has "mtr-packet" as a hard-coded executable name.
This could possibly be abstracted to a config.h `#define` as part of the configure script.

Komzpa:
## Summary

- apply the configured program transform to the installed `mtr-packet` privilege hook
- define the transformed packet helper name for runtime lookup when `--program-prefix` or `--program-suffix` is used
- keep the untransformed `./mtr-packet` build-tree fallback for local testing/debugging

Fixes #191.

## Validation

- `git diff --check`
- `./bootstrap.sh`
- `./configure --without-gtk --without-jansson --program-suffix=-0.99-test`
- verified `config.h` defines `MTR_PACKET_NAME "mtr-packet-0.99-test"`
- verified `make -n install-exec-hook` applies the configured make transform
- `make -j "$(nproc)"` with suffix configure
- `./mtr --report --report-cycles 1 -m 1 127.0.0.1` with suffix configure
- `make -j "$(nproc)"`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,8 +41,9 @@ dist_man_MANS = mtr.8 mtr-packet.8
 PATHFILES += man/mtr.8 man/mtr-packet.8
 
 install-exec-hook:
-	`setcap cap_net_raw+ep $(DESTDIR)$(sbindir)/mtr-packet` \
-	|| chmod u+s $(DESTDIR)$(sbindir)/mtr-packet
+	mtr_packet_name=`echo mtr-packet | sed '$(transform)'`; \
+	`setcap cap_net_raw+ep $(DESTDIR)$(sbindir)/$$mtr_packet_name` \
+	|| chmod u+s $(DESTDIR)$(sbindir)/$$mtr_packet_name
 
 mtr_SOURCES = ui/mtr.c ui/mtr.h \
               ui/net.c ui/net.h \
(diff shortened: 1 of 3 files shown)


Komzpa:
Superseded by #578, which keeps this fix together with the related install-hook setcap fallback so the merge stack no longer has an internal Makefile.am conflict.

PR #584, opened 2026-05-07
curses: add report snapshot on exit

### LINKED ISSUE #323: Don't clean screen after quit
I see that when I exit the program it will clear the screen, losing all the information displayed.

It would be nice if it kept the trace information (or that it would print it again if required), so it can be read or copied later.

Komzpa:
## Summary

- add an opt-in `--report-on-exit` flag for the curses interface
- close curses normally, then print a standard report snapshot so the last trace remains visible in terminals that use an alternate screen
- document the new option in `mtr(8)` and `--help`

This follows the workaround discussed in #323 without changing the default curses behavior for existing users.

Fixes #323.

## Validation

- `git diff --check`
- `./bootstrap.sh`
- `make -j "$(nproc)"`
- `./mtr --help 2>&1 | grep -F -- --report-on-exit`
- `script -q -e -c "sh -c 'sleep 1; printf q' | ./mtr --report-on-exit --curses --no-dns --max-ttl 1 localhost" /tmp/mtr-report-on-exit.typescript`

### DIFF
--- a/man/mtr.8.in
+++ b/man/mtr.8.in
@@ -194,6 +194,11 @@ mode.  When in this mode,
 .B mtr
 will not cut hostnames in the report.
 .TP
+.B \-\-report\-on\-exit
+Use this option with the curses interface to print a report snapshot after
+leaving curses mode. This keeps the last trace visible in terminals that switch
+to an alternate screen while curses is active.
+.TP
 .B \-x\fR, \fB\-\-xml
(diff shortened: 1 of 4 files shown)


Komzpa:
Superseded by #580, which now carries both curses startup/exit controls so the merge stack no longer has an internal option-plumbing conflict.

PR #598, opened 2026-05-07
fix(cmdpipe): bound probe command appends

Komzpa:
## Summary

Build appended `send-probe` command fragments directly into the remaining command buffer through one checked `append_command_format()` helper.

This removes the temporary argument buffers plus `strncat()` pattern, avoids converting a negative remaining length into a large unsigned append size if the command is already at the buffer boundary, and turns command-buffer truncation into a clear `mtr-packet command too long ...` failure instead of silently sending a partial command to `mtr-packet`.

The `mtr-packet` command names, argument names, and protocol names used by this path are now defined in `packet/cmdparse.h` so the UI command builder does not keep repeating wire-format string literals.

The command format is unchanged when it fits.

## Validation

- `make -j$(nproc)`
- `git diff --check`
- `./mtr -r -c 1 127.0.0.1`

Tracked issues closed: none. This is analyzer/string-safety cleanup.

### DIFF
--- a/packet/cmdparse.h
+++ b/packet/cmdparse.h
@@ -24,6 +24,28 @@ enum {
     MAX_COMMAND_TOKENS = MAX_COMMAND_ARGUMENTS * 2 + 2
 };
 
+#define COMMAND_NAME_SEND_PROBE "send-probe"
+
+#define COMMAND_ARG_BIT_PATTERN "bit-pattern"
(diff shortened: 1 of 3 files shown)


rewolff:
I fixed a bunch of these (in case the compiler issued a warning). Lets revisit in a day or two, ok ?

PR #603, opened 2026-05-07
fix(dns): harden resolver lifecycle and results

### LINKED ISSUE #220: MTR dns-lookup process does not exit with the rest.
The dns lookup process does not exit when you hit "q" in the terminal output version. (not tested in QT version). 
It then reports  `./mtr: write DNS lookup result: Broken pipe` a few seconds later. 

I think it was still doing a lookup so if it had not been doing that, it probably would have kept on running (not noticing it's input pipe was gone? Hmm. 

(To trigger: Have an address that takes a while to resolve, and quit before it happens). 

TODO: Keep a list of DNS processes and kill (send a signal) them when we exit. Or (simpler, but error-prone) just ignore that one error: "broken pipe" on writing the result. The problem with that last approach is that when this error does occur unexpectedly we might ignore it and not provide any feedback as to why things are failing.

### LINKED ISSUE #448: Completion of error handling
Would you like to add more error handling for return values from functions like the following?
- [fdopen](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopen.html "Associate a stream with a file descriptor") ⇒ [dns_open](https://github.com/traviscross/mtr/blob/ec42ba61f77654e8397e6496095634585f90b26d/ui/dns.c#L150)
- [printf](https://pubs.opengroup.org/onlinepubs/9699919799/functions/printf.html "Print formatted output") ⇒ [xml_close](https://github.com/traviscross/mtr/blob/0886eee5bab164fee50f94d542de71fca8a2dc68/ui/report.c#L403)

### LINKED ISSUE #449: Problems with hosts that have a NULL PTR result
I've noticed that if a host has a null result to a PTR lookup, it seems that the resultant host output isn't properly initialised / terminated.

This isn't one of my domains, I just stumbled across it.

% host  188.209.56.29 
29.56.209.188.in-addr.arpa domain name pointer .

See pasted output below.
With mtr v0.95 installed, I get the first output.

The second output is from mtr from github master, checked out 10 minutes ago.

Note the "host" displayed in the second case is actually the value of the host in hop 7.

The third output is done via ip6 - note again how the host in this case is actually the host from hop 6.

I've tried various other combinations, and get similar results.

2001:470:1f07:10f1::3:5 is one of mine - I set its ptr value to "." to emulate the problem.

Let me know if I can help further!

Cheers, Jamie

```
$ mtr -b 188.209.56.29
thompson.dyslexicfish.net (172.20.1.1) -> 188.209.56.29 (188.209.56.29)
Keys:  Help   Display mode   Restart statistics   Order of fields   quit

 Host
 1. router.home.dyslexicfish.net (172.20.0.1)
 2. 88-109-32-1.dynamic.dsl.as9105.com (88.109.32.1)
 3. ae52-ner002.bir.as13285.net (78.144.1.31)
 4. ae52-scr102.loh.as13285.net (78.144.1.30)
 5. ae62-scr101.thw.as13285.net (78.144.1.114)
 6. linx.r1.gsa.nl.as43350.com (195.66.226.121)
 7. 185.107.116.3 (185.107.116.3)
 8. �����^? (188.209.56.29)

$ ./mtr -b 188.209.56.29

thompson.dyslexicfish.net (172.20.1.1) -> 188.209.56.29 (188.209.56.29)
Keys:  Help   Display mode   Restart statistics   Order of fields   quit

 Host
 1. router.home.dyslexicfish.net (172.20.0.1)
 2. 88-109-32-1.dynamic.dsl.as9105.com (88.109.32.1)
 3. ae52-ner002.bir.as13285.net (78.144.1.31)
 4. ae52-scr102.loh.as13285.net (78.144.1.30)
 5. ae62-scr101.thw.as13285.net (78.144.1.114)
 6. linx.r1.gsa.nl.as43350.com (195.66.226.121)
 7. 185.107.116.3 (185.107.116.3)
    185.107.116.4 (185.107.116.4)
 8. 185.107.116.3 (188.209.56.29)

$ ./mtr -b 2001:470:1f07:10f1::3:5

thompson.dyslexicfish.net (2001:470:1f1d:bb::1:1) -> 2001:470:1f07:10f1::3:5
Keys:  Help   Display mode   Restart statistics   Order of fields   quit

 Host
 1. router.home.dyslexicfish.net (2001:470:1f1d:bb::1)
 2. tunnel282020.tunnel.tserv1.lon2.ipv6.he.net (2001:470:1f1c:bb::1)
 3. e0-18.core2.lon2.he.net (2001:470:0:320::1)
 4. (waiting for reply)
 5. 100ge4-1.core1.nyc4.he.net (2001:470:0:2cf::2)
 6. tserv1.nyc4.he.net (2001:470:0:5d::2)
 7. tserv1.nyc4.he.net (2001:470:1f07:10f1::3:5)

```

Komzpa:
Fixes #220.
Fixes #448.
Fixes #449.
Supersedes #608.
Supersedes #609.

## Summary
- terminate DNS resolver subprocesses on exit instead of leaving slow lookups behind
- ignore empty and single-dot PTR hostnames so output falls back to numeric addresses
- fail explicitly if DNS resolver pipe setup fails (`fdopen()`/`fcntl()`)

## Validation
- `make -j$(nproc)`
- `git diff --check`
- `./mtr -r -c 1 127.0.0.1`
- earlier focused validation for the original pieces covered curses quit/no-leftover-process checks and the local empty-PTR resolver case.

### DIFF
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -114,6 +114,13 @@ static void set_sockaddr_ip(
     memcpy(sockaddr_addr_offset(sa), ip, sockaddr_addr_size(sa));
 }
 
+static int is_useful_hostname(
+    const char *hostname)
+{
+    return hostname[0] != '\0'
+        && !(hostname[0] == '.' && hostname[1] == '\0');
+}
+
 void dns_open(
     void)
 {
@@ -176,7 +183,7 @@ void dns_open(
 
                 rv = getnameinfo((struct sockaddr *) &sa, salen,
                                  hostname, sizeof(hostname), NULL, 0, 0);
-                if (rv == 0) {
+                if (rv == 0 && is_useful_hostname(hostname)) {
                     snprintf(result, sizeof(result),
                              "%s %s\n", strlongip(family, &host), hostname);
 
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -162,6 +162,9 @@ void dns_open(
             close(i);
         }
         infp = fdopen(todns[0], "r");
+        if (!infp) {
+            error(EXIT_FAILURE, errno, "fdopen DNS lookup pipe");
+        }
 
         while (fgets(buf, sizeof(buf), infp)) {
             ip_t host;
@@ -207,9 +210,17 @@ void dns_open(
         close(todns[0]);        /* close the pipe ends we don't need. */
         close(fromdns[1]);
         fromdnsfp = fdopen(fromdns[0], "r");
+        if (fromdnsfp == NULL) {
+            error(EXIT_FAILURE, errno, "fdopen DNS result pipe");
+        }
         flags = fcntl(fromdns[0], F_GETFL, 0);
+        if (flags < 0) {
+            error(EXIT_FAILURE, errno, "fcntl DNS result pipe");
+        }
         flags |= O_NONBLOCK;
-        fcntl(fromdns[0], F_SETFL, flags);
+        if (fcntl(fromdns[0], F_SETFL, flags) < 0) {
+            error(EXIT_FAILURE, errno, "fcntl DNS result pipe");
+        }
     }
 }
 
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -36,6 +36,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <signal.h>
+#include <sys/wait.h>
 
 #include "mtr.h"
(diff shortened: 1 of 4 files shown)


rewolff:
I would really like a followup pull where the two copies of "is_useful_hostname"  are merged. How about static inline in a header somewhere?

PR #605, opened 2026-05-07
docs: clarify how mtr probes hops

### LINKED ISSUE #239: Documentation/description error?
I might be very thick today, but there seems to be an error in the description. 

It says:
 _As mtr starts, it investigates the network connection between the host
  mtr runs on and a user-specified destination host.  After it
  determines the address of each network hop between the machines, 
  it sends a sequence of ICMP ECHO requests to each one to determine the 
  quality of the link to each machine.  As it does this, it prints
  running statistics about each machine._

What I am seeing actually is a constant traceroute-like behavior, sending packets with the final IP address as a destination and increasing TTL. I am trying and it is *not* sending any packets to any network hops. All of the packets go to the final destination.

So, please, update the description in order to make it more accurate. From that description a user can infer that "packet loss" from one of the intermediate routers means that there is actual packet loss in the link, while it actually means that some rate limit mechanism is operating in the control plane filtering the ICMP TTL exceeded packets.

Thanks!

Komzpa:
## Summary

Fixes #239.

The README description said that after mtr determines each network hop, it sends ICMP ECHO requests to each one.  That can mislead readers into treating apparent loss at an intermediate hop as direct evidence of forwarded packet loss there.

This rewords the description to say that mtr sends probes to the destination with increasing TTL values, uses responses from intervening routers to discover and measure the path, and notes that intermediate routers may be configured not to send those ICMP responses or may rate-limit them.

## Validation

- `git diff --check origin/master...HEAD`
- `./bootstrap.sh && ./configure --without-gtk --without-jansson`
- `make -j "$(nproc)"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,15 @@ mtr combines the functionality of the 'traceroute' and 'ping' programs
 in a single network diagnostic tool.
 
 As mtr starts, it investigates the network connection between the host
-mtr runs on and a user-specified destination host.  After it
-determines the address of each network hop between the machines,
-it sends a sequence of ICMP ECHO requests to each one to determine the
-quality of the link to each machine.  As it does this, it prints
-running statistics about each machine.
+mtr runs on and a user-specified destination host.  It sends probes to
+the destination with successively larger time-to-live values, and uses
+the responses from intervening routers to discover and measure the path.
+As it does this, it prints running statistics for each hop.
+
+Intermediate routers may be configured to never send these ICMP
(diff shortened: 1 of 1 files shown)


rewolff:
What I see more often than rate limit and stuff like that is that they are configured to simply never send the replies. 

How about: 

> Intermediate routers may be configured to never send these ICMP responses or impose a rate-limit on them, so apparent loss at an ...

PR #606, opened 2026-05-07
refactor(csv): remove unused close loop

### LINKED ISSUE #455: Delete questionable loop at the beginning of csv_close()
I noticed [the following loop in the implementation of the function “csv_close”](https://github.com/traviscross/mtr/blob/0886eee5bab164fee50f94d542de71fca8a2dc68/ui/report.c#L478 "Update candidate").

```c
    for (i = 0; i < MAXFLD; i++) {
        j = ctl->fld_index[ctl->fld_active[i]];
        if (j < 0)
            continue;
    }
```

💭 I find such a control flow not useful.
Thus I suggest to remove it.

Komzpa:
## Summary

Fixes #455.

`csv_close()` had a leading loop over the active fields that only assigned `j` and continued on invalid fields.  It did not produce output, validate state, or affect later control flow.  The real header and row generation loops immediately below still perform the same field filtering.

This removes the no-op loop.

## Validation

- `make -j "$(nproc)"`
- `git diff --check`
- `./mtr -C -c 1 127.0.0.1`

### DIFF
--- a/ui/report.c
+++ b/ui/report.c
@@ -474,12 +474,6 @@ void csv_close(
     ip_t *addr2 = NULL;
     char name[MAX_FORMAT_STR];
 
-    for (i = 0; i < MAXFLD; i++) {
-        j = ctl->fld_index[ctl->fld_active[i]];
-        if (j < 0)
-            continue;
-    }
-
     max = net_max(ctl);
     at = net_min(ctl);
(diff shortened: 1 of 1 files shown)


rewolff:
I like patches that reduce the total lines-of-code! 

:-)

PR #609, opened 2026-05-07
fix(dns): check resolver pipe setup failures

### LINKED ISSUE #448: Completion of error handling

Komzpa:
Fixes #448.

## Summary
- fail explicitly if DNS resolver pipe `fdopen()` setup fails in the helper or parent
- check both `fcntl()` calls before using the resolver result pipe in nonblocking mode

## Validation
- `make -j$(nproc)`
- `git diff --check`
- `./mtr --version`
- `./mtr --help`
- `./mtr -r -c 1 127.0.0.1`

## Notes
This intentionally addresses the concrete unchecked pipe setup path from #448 without adding broad output-wrapper macros around `printf()`/`putchar()`.

### DIFF
--- a/ui/dns.c
+++ b/ui/dns.c
@@ -148,6 +148,9 @@ void dns_open(
             close(i);
         }
         infp = fdopen(todns[0], "r");
+        if (infp == NULL) {
+            error(EXIT_FAILURE, errno, "fdopen DNS input pipe");
+        }
 
         while (fgets(buf, sizeof(buf), infp)) {
             ip_t host;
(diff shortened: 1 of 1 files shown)


Komzpa:
Superseded by the updated #603, which now carries this DNS pipe setup hardening together with the related resolver lifecycle/result fixes.

PR #615, opened 2026-05-08
fix(install): honor transformed packet helper

### LINKED ISSUE #191: Issue with "--program-suffix" option to ./configure

### LINKED ISSUE #203: setcap command not found
Cygwin on win 10.
./bootstrap.sh
./configure && make
make install

make  install-exec-hook
make[2]: Entering directory '/home/user/Downloads/mtr'
`setcap cap_net_raw+ep /usr/local/sbin/mtr-packet` \
|| chmod u+s /usr/local/sbin/mtr-packet
/bin/sh: line 1: setcap: command not found
make[2]: Leaving directory '/home/user/Downloads/mtr'
 /usr/bin/mkdir -p '/usr/local/share/bash-completion/completions'
 /usr/bin/install -c -m 644 bash-completion/mtr '/usr/local/share/bash-completion/completions'
 /usr/bin/mkdir -p '/usr/local/share/man/man8'
 /usr/bin/install -c -m 644 mtr.8 mtr-packet.8 '/usr/local/share/man/man8'
make[1]: Leaving directory '/home/user/Downloads/mtr'

Komzpa:
Fixes #191.
Fixes #203.
Supersedes #578.
Supersedes #583.

## Summary
- install capabilities/setuid fallback on the transformed `mtr-packet` program name
- avoid noisy `setcap: command not found` output by checking whether `setcap` is available before using it
- teach runtime lookup to prefer the transformed packet helper name while keeping `MTR_PACKET` and `./mtr-packet` debugging fallback support

## Validation
- `./bootstrap.sh`
- `./configure --program-prefix=x --prefix=/tmp/mtr-transform-combined`
- `make -j$(nproc)`
- `git diff --check`
- `PATH=/usr/bin:/bin make install DESTDIR=/tmp/mtr-transform-install` installs `xmtr-packet` without setcap lookup noise and applies setuid fallback
- `/tmp/mtr-transform-install/tmp/mtr-transform-combined/sbin/xmtr -r -c 1 127.0.0.1`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,8 +41,13 @@ dist_man_MANS = mtr.8 mtr-packet.8
 PATHFILES += man/mtr.8 man/mtr-packet.8
 
 install-exec-hook:
-	`setcap cap_net_raw+ep $(DESTDIR)$(sbindir)/mtr-packet` \
-	|| chmod u+s $(DESTDIR)$(sbindir)/mtr-packet
+	mtr_packet_name=`echo mtr-packet | sed '$(transform)'`; \
+	if command -v setcap >/dev/null 2>&1; then \
+		setcap cap_net_raw+ep $(DESTDIR)$(sbindir)/$$mtr_packet_name \
+		|| chmod u+s $(DESTDIR)$(sbindir)/$$mtr_packet_name; \
+	else \
+		chmod u+s $(DESTDIR)$(sbindir)/$$mtr_packet_name; \
+	fi
 
 mtr_SOURCES = ui/mtr.c ui/mtr.h \
               ui/net.c ui/net.h \
--- a/configure.ac
+++ b/configure.ac
@@ -287,6 +287,22 @@ AC_ARG_ENABLE([bash-completion],
   [], [enable_bash_completion=yes]
 )
 AM_CONDITIONAL([BUILD_BASH_COMPLETION], [test "x$enable_bash_completion" = xyes])
+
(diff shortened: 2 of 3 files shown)


Komzpa:
No longer needed: the original PR #578 has been updated in-place with this combined fix.

PR #616, opened 2026-05-08
feat(curses): add startup and exit controls

### LINKED ISSUE #323: Don't clean screen after quit

### LINKED ISSUE #549: Feature: command line argument to start in compact view
## Description

There is no option to start in compact view. I would like this option.

## User story

Sally wants to write a script that shows the compact view of mtr, without reverting to having to send interactive keystrokes programmatically.

## Version info

Tested on `mtr` v0.96

## Additional info

Curiously, when running `mtr` and pressing `h` to view the available commands, `c` is listed twice:
```
Command:
...
  c       switching compact mode
...
  c <n>   report cycle n, default n=infinite
```
The second listing does not appear to work when using interactive mode though, but is a command line argument.

Komzpa:
Fixes #323.
Fixes #549.
Supersedes #580.
Supersedes #584.

## Summary
- add `--compact` to start the curses UI in compact layout
- add `--report-on-exit` to print a report snapshot after leaving curses mode
- document both options and include them in bash completion

## Validation
- `make -j$(nproc)`
- `git diff --check`
- `./mtr --help | rg -- '--compact|--report-on-exit'`
- `./mtr --compact --report-on-exit -r -c 1 127.0.0.1`

### DIFF
--- a/bash-completion/mtr
+++ b/bash-completion/mtr
@@ -55,7 +55,7 @@ _mtr_module()
         --max-ttl --max-unknown --port --localport --psize --bitpattern
         --interval --gracetime --tos --mpls --timeout --mark --report
         --report-wide --report-cycles --json --xml --csv --raw --split
-        --curses --displaymode --gtk --no-dns --show-ips --order --ipinfo
+        --curses --displaymode --compact --report-on-exit --gtk --no-dns --show-ips --order --ipinfo
         --aslookup --help --version
       '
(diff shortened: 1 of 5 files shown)


Komzpa:
No longer needed: the original PR #580 has been updated in-place with this combined fix.

PR #620, opened 2026-05-08
net: resolve IDN hostnames with AI_IDN

Komzpa:
## Summary

Cherry-picked a small, dependency-free part of the IDN support from `yvs2014/mtr085`.

When libc exposes `AI_IDN`, pass it to `getaddrinfo()` for target resolution and multi-target report preflight.  This lets libc convert internationalized hostnames such as `köthe.de` to their DNS form without adding libidn/libidn2 as a new dependency.

Platforms without `AI_IDN` keep the existing behavior because configure records `HAVE_DECL_AI_IDN=0`.

## Provenance

- Ported from `yvs2014/mtr085` commit `f897e61ad950499acdf213207fe08b5dce75293e` (`rm libidn dependenccy if IDN is supported by getaddrinfo`).
- Original author: yvs `<VSYakovetsky@gmail.com>`.
- This PR ports only the libc `AI_IDN` part, not the libidn/libidn2 fallback.

## Validation

- `./bootstrap.sh && ./configure --without-gtk --without-jansson`
- `git diff --check`
- `make -j "$(nproc)"`
- `timeout 35s env MTR_PACKET=./mtr-packet ./mtr -r -c 1 köthe.de`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -203,6 +203,13 @@ AC_CHECK_DECLS([errno], [], [], [[
 #include <errno.h>
 #include <sys/errno.h>
   ]])
+AC_CHECK_DECLS([AI_IDN], [], [], [[
+#include <sys/types.h>
+#ifdef HAVE_SYS_SOCKET_H
+#include <sys/socket.h>
+#endif
+#include <netdb.h>
+  ]])
 
 AC_CHECK_TYPE([socklen_t],
   [AC_DEFINE([HAVE_SOCKLEN_T], [], [Define if your system has socklen_t])], [],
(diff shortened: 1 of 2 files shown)


yvs2014:
AI_IDN can be tested in a bit simpler way, i.e.
```c
#ifdef AI_IDN
```
instead of
```c
#if HAVE_DECL_AI_IDN
```
like it's here https://github.com/yvs2014/mtr085/blob/9dc18628b35aa347a0c81318ecf633a33b39f3e8/mtr.c#L1111

And to be sure in the correct header inclusion for macro checks, something like (for autotools build system)
`AC_CHECK_HEADERS([netdb.h], AC_DEFINE(HAVE_NETDB_H, 1))` in configure.ac
and then
```c
#ifdef HAVE_NETDB_H
#include <netdb.h>
#endif
```

p.s. it's the way for glibc case, for other libc-es it needs to work directly with libidn

PR #628, opened 2026-05-08
curses: validate packet size input

Komzpa:
## Summary

Port the interactive packet-size validation piece from `yvs2014/mtr085`.

The curses `s` command now validates the newly entered packet size before assigning it. This keeps the interactive path aligned with command-line `--psize`, including negative random packet-size values.

## Provenance

- Ported from `yvs2014/mtr085` commit `e100fd72be37f0a98d6893db7e0baa8a5b81b72a` (`use arc4random if it's available, and put extra checks for -B/-s options`).
- Original author: yvs `<VSYakovetsky@gmail.com>`.
- Commit author is preserved as yvs; this PR ports only the curses packet-size validation piece.

## Validation

- `git diff --check`
- `make -j "$(nproc)"`

### DIFF
--- a/ui/curses.c
+++ b/ui/curses.c
@@ -217,8 +217,11 @@ int mtr_curses_keyaction(
         }
         buf[i] = '\0';
         int new_packetsize = atoi(buf);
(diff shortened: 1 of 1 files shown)


rewolff:
Good catch.

PR #642, opened 2026-05-08
test(output): cover bad host exit status

### LINKED ISSUE #294: Return exit code when host not found
When mtr is used on the command line and an invalid host is passed, mtr raises the message "Failed to resolve host: Name or service not known" but it returns 0, not -1 or any other code. Returning something different from zero when exiting the process with invalid host would make using it with other commands and pipes easier.

Komzpa:
## Summary
- add regression coverage for failed hostname resolution in non-interactive output modes
- verify `--csv`, `--raw`, `--report`, `--xml`, and `--json` when available all return non-zero and print the resolver error

Fixes #294.

The current code already exits non-zero in these cases; this keeps the historical `--csv`/`--raw`/`--json` failure mode covered.

## Tests
- `git diff --check`
- `./bootstrap.sh && ./configure --without-gtk --without-jansson && make -j "$(nproc)"`

### DIFF
--- a/bash-completion/mtr
+++ b/bash-completion/mtr
@@ -54,7 +54,8 @@ _mtr_module()
         --filename --inet --inet6 --udp --tcp --address --first-ttl
         --max-ttl --max-unknown --port --localport --psize --bitpattern
         --interval --gracetime --tos --mpls --timeout --mark --report
-        --report-wide --report-cycles --json --xml --csv --raw --split
+        --report-wide --report-on-exit --report-cycles --json --xml --csv
+        --raw --split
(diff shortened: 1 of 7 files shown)


rewolff:
FYI: You're issuing patches at a higher rate than what I can
check-and-approve. So I have to prioritize a few other things now.

Komzpa:
@rewolff please do not be stressed about it, there is no rush, and thanks for taking the time to review and merge the things you merged already. I hope the patches are useful; but if ticket was there since 2017 it can probably wait a bit longer so everyone can have a balanced life.

